PR #3, opened 2016-08-15
Standardized Readme

RichardLitt:
See multiformats/multiformats#13

What about the Install section?


dignifiedquire:
THanks :octocat:

PR #4, opened 2016-09-11
Remove the `drop_with_repr_extern` lint.

eminence:
This lint was removed in nightly rust.  Note that while this lint still
exists in the Stable and Beta channels, it only applies if we have types
with `#[repr(C)]` that implement `Drop`.  Since this library has neither
(and will probably never have them), it is safe to remove this lint.


dignifiedquire:
Thanks :octocat:

PR #6, opened 2016-09-30
Support multiple hash function.

zypeh:
Changed from `sodiumoxide` cargo package to `rust-crypto` which is more active development and well-documented.

Now should be support:
- `SHA1`
- `SHA3`
- `SHAKE`
- `Blake2b` & `Blake2s`

Will update the readme and docs in future.

### DIFF
--- a/src/hashes.rs
+++ b/src/hashes.rs
@@ -9,6 +9,8 @@ pub enum HashTypes {
     SHA2256,
     /// SHA2-512 (64-byte hash size)
     SHA2512,
+    /// SHA3-512 (64-byte hash size)
+    SHA3512,
     /// SHA3-384 (48-byte hash size)
     SHA3384,
     /// SHA3-256 (32-byte hash size)
@@ -53,6 +55,8 @@ impl HashTypes {
         HashTypes::SHA3384  => 48,
         HashTypes::SHA3256  => 32,
         HashTypes::SHA3224  => 28,
+        HashTypes::SHAKE128 => 16,
+        HashTypes::SHAKE256 => 32,
 	    HashTypes::Blake2b  => 64,
 	    HashTypes::Blake2s  => 32,
         }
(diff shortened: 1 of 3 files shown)


src/hashes.rs
```diff
@@ -48,7 +68,12 @@ impl HashTypes {
 	    HashTypes::SHA1     => "SHA1",
 	    HashTypes::SHA2256  => "SHA2-256",
 	    HashTypes::SHA2512  => "SHA2-512",
-	    HashTypes::SHA3     => "SHA3",
+	    HashTypes::SHA3512  => "SHA3-512",
```
dignifiedquire: Identation is off here

src/lib.rs
```diff
@@ -53,11 +60,91 @@ pub use hashes::*;
...
+            hasher.result(&mut buf);
+            buf
+        },
+
+        HashTypes::Blake2s  => {
+            let mut buf : Vec<u8> = Vec::new();
+            let mut hasher = Blake2s::new(32);
+            hasher.input(input);
+            hasher.result(&mut buf);
+            buf
```
dignifiedquire: All the methods above look pretty much the same, could you refactor them to share a little bit more code?

zypeh:
Ops, my bad, I didn't write tests. Will fix that.

src/lib.rs
zypeh: Sure :) but let me catch some sleep first 👍

dignifiedquire: Gotta catchem all ;)

chpio: you could also preallocate the vec:
```rust
let mut buf : Vec<u8> = Vec::with_capacity(wanttype.size());
```

ps: (IMHO) HashTypes.size should return *usize*

dignifiedquire:
Closing in favor of #8

PR #12, opened 2017-02-06
Keccak + SHA3

maciejhirsz:


Cargo.toml
```diff
@@ -5,7 +5,7 @@ homepage = "https://github.com/multiformats/rust-multihash"
 
 keywords = ["multihash", "ipfs"]
 
-version = "0.4.2"
+version = "0.4.3"
```
dignifiedquire: Can you remove this? I like to make version bumps its own commit, so I can tag it.

tests/lib.rs
```diff
@@ -14,45 +14,55 @@ fn hex_to_bytes(s: &str) -> Vec<u8> {
...
-        Hash::SHA2256
-    );
-
-    let hash: Vec<u8> = encode(Hash::SHA2512, b"helloworld").unwrap();
-    assert_eq!(
-        decode(&hash).unwrap().alg,
-        Hash::SHA2512
+    assert_decode!(
+        SHA1, SHA2256, SHA2512, SHA3224, SHA3256, SHA3384, SHA3512,
+        Keccak224, Keccak256, Keccak384, Keccak512
```
dignifiedquire: while you are in here, could you add a roundtrip macro test as well please?

maciejhirsz: decoding was already a roundtrip test, added new ones for decoding from a predefined hashes instead

dignifiedquire:
🎉

PR #14, opened 2017-03-17
Bump ring

tomusdrw:
Current version does not compile on latest stable rustc.
ping @dignifiedquire


dignifiedquire:
Has parity upgraded as well? I know that @maciejhirsz mentioned having issues before with non matching ring versions.

maciejhirsz:
@dignifiedquire The conflicting ring dependency has been since removed from Parity.

dignifiedquire:
Thank you

PR #16, opened 2017-04-13
Replace "homepage" link with "repository" link.

briansmith:
The crates.io page at https://crates.io/crates/multihash makes it look
like there's no public repository for the project. Improve this by
replacing the "homepage" link with a "repository" link.


dignifiedquire:
Thank you :octocat:

PR #19, opened 2017-11-02
Upgrade ring to 0.12

tomaka:
For exactly the same reason as #18, this PR upgrades the ring library to version 0.12.
It also bumps multihash to 0.7.0.


dignifiedquire:
thanks, can you remove the version bump on both prs for the lib itself, I want to keep those in their own commits for tagging

tomaka:
Done, I opened #20 instead.

PR #21, opened 2017-11-24
Implement Hash for Hash

tomaka:


dignifiedquire:
Thanks

tomaka:
Actually it turns out this change is not needed for libp2p, but it's probably a nice thing to have anyway.

PR #23, opened 2018-01-19
Fix a panic when decoding empty input

tomaka:


tomaka:
Self-merging because this is trivial.

PR #26, opened 2018-02-21
Rename `encode` to `sum`, introduce new `encode`

### LINKED ISSUE #25: No way to `encode` a digest directly?
# Problem

If you have a hash digest `digest: &[u8]` and you want to `encode` that into multihash bytes, you're out of luck if you're using `rust-multihash` currently. This is because `encode` expects un-hashed data as input. Instead, you must do something like this manually:

```
    let hash_alg = multihash::Hash::SHA1; 
    let mut mh = Vec::with_capacity(digest.len() + 2); 
    mh.push(hash_alg.code()); 
    mh.push(hash_alg.size()); 
    mh.extend_from_slice(digest); 
```

In contrast, `go-multihash`'s `Encode` function does not hash your data for you, so you can directly pass in a digest.

# Potential solutions

This could be remedied with a new encoding function that expects a hash digest, maybe something named `encode_digest`? However, perhaps we should consider changing the way `encode` works. It would be a breaking change, but it would bring it in-line with how go-multihash works.

leavehouse:
In go-multihash, the `Sum` function takes un-hashed data and a
specification of a hash function and produces multihash bytes
by hashing the input data. The `Encode` function, on the other hand,
produces multihash bytes from an existing hash digest.

This commit adds an equivalent of the `Encode` function from
go-multihash. Since there is already a function named "encode",
this commit also renames it to `sum` to make it similar to the go
version of this library. This is a breaking change.

Closes #25

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -61,16 +61,16 @@ macro_rules! match_encoder {
 /// # Examples
 ///
 /// ```
-/// use multihash::{encode, Hash};
+/// use multihash::{sum, Hash};
 ///
 /// assert_eq!(
-///     encode(Hash::SHA2256, b"hello world").unwrap(),
+///     sum(Hash::SHA2256, b"hello world").unwrap(),
 ///     vec![18, 32, 185, 77, 39, 185, 147, 77, 62, 8, 165, 46, 82, 215, 218, 125, 171, 250, 196,
 ///     132, 239, 227, 122, 83, 128, 238, 144, 136, 247, 172, 226, 239, 205, 233]
 /// );
 /// ```
 ///
-pub fn encode(hash: Hash, input: &[u8]) -> Result<Vec<u8>, Error> {
+pub fn sum(hash: Hash, input: &[u8]) -> Result<Vec<u8>, Error> {
     let size = hash.size();
     let mut output = Vec::new();
     output.resize(2 + size as usize, 0);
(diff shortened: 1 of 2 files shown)


leavehouse:
I know the testing for the new `encode` function isn't adequate, but since this is a breaking change I wanted to get feedback before fleshing that out.

The alternative to this breaking change is to not try to have it similarly named to the functions in go-multihash, and simply give this function a new name. `encode_digest` is one option I've thought of that seems decent.

tomaka:
I know that naming is subjective, but for me what makes sense would be the opposite: sum takes some already-hashed data, and encode performs the hashing.

leavehouse:
I like "sum" for describing "pushing data through a hash function", especially since it lines up with the names of existing command line programs for hashing data like `sha1sum`, `sha256sum`, etc.

I think it's also good to point out that `js-multihash`'s encode function does not perform hashing, and instead [takes a digest](https://multiformats.github.io/js-multihash/#multihashencode). So this name change would bring rust-multihash in-line with both go-multihash and js-multihash.

crackcomm:
I think that makes sense, `sum` to compute the hash and `encode` and `decode` for multihash encoding.

It also follows logic used in Go (see `hash.Sum([]byte) ([]byte, error)`).

@leavehouse pointed to js-multihash, same is true for go-multihash [`func Sum(data []byte, code uint64, length int) (Multihash, error)
`](https://godoc.org/github.com/multiformats/go-multihash#Sum).

> Sum obtains the cryptographic sum of a given buffer.

[`func Encode(buf []byte, code uint64) ([]byte, error)`](https://godoc.org/github.com/multiformats/go-multihash#Encode)

> Encode a hash digest along with the specified function code.

It's not a subjective matter any longer.

dignifiedquire:
I want to follow the Digest trait that rust crypto established for now. https://docs.rs/digest/0.8.0/digest/ (see the work in #30)

For the explicit step of taking a raw hash and wrapping it, I don’t have a strong preference

vmx:
If you want to encode a Multihash without doing any hashing, you can now use `multihash::wrap()`.

PR #30, opened 2018-11-06
[wip] next version

dignifiedquire:
- add support for blake2b and blake2s
- add support for codes larger than 255
- add support for streaming based hashing
- add `std` feature, and test for `no_std` compat (needs verification)
- add `asm` and `simd` features to allow enabling those on the underlying hashers
- minimize overhead of `Multihash` wrapper type around the actual hash
- update travis script and ensure features are being tested
- use proc_macro to do the actual implementation under the hood


src/digests.rs
```diff
@@ -0,0 +1,213 @@
...
+#[derive(Clone, Debug, PartialEq, MultihashDigest)]
+#[repr(u32)]
+pub enum Code {
+    #[Size = "20"]
+    #[Digest = "sha1::Sha1"]
+    Sha1 = 0x11,
+
+    #[Size = "32"]
+    #[Digest = "sha2::Sha256"]
+    Sha2256 = 0x12,
```
crackcomm: Does it really have to be like that?

dignifiedquire: which part?

crackcomm: I mean the `Sha2256`. I understand need for no `_` but something like `Sha256V2`,  `Sha256V3`, `Sha512V2`. Anything to make it valid syntactically.

multihash_derive/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+        .expect("Please supply a Size attribute")
+        .parse()
+        .expect("Size should be a number");
+
+    if size > 255 {
+        panic!("Sizes larger than 255 are not yet supported");
+    }
+
+    let code =
+        fetch_discriminant(&variant.discriminant).expect("Please provide the code as discriminant");
```
laser: so very polite!

multihash_derive/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+            fn to_string() -> &'static str {
+                stringify!(#name)
+            }
+
+            fn code() -> Code {
+                Code::#name
+            }
+
+            fn wrap(
+                raw: &[u8],
```
laser: I notice that the `input` method, above, accepts `<AsRef[u8]>`. 

This function accepts `&[u8]. 

I suggest that you use `AsRef<[u8]>`, for consistency.

multihash_derive/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+        impl BlockInput for #name {
+            type BlockSize = <#original_digest as BlockInput>::BlockSize;
+        }
+
+        impl MultihashDigest for #name {
+            fn new() -> Self {
+                #name(<#original_digest as Digest>::new())
+            }
+
+            fn size() -> usize {
```
laser: Honest question: Why does this function return `usize` instead of `u8`? I see that you're doing some runtime checks to ensure that all `#size` values are in the range of unsigned, 8-bit integers anyways.

multihash_derive/src/lib.rs
crackcomm: Not just consistency though, good point.

src/digests.rs
crackcomm: @dignifiedquire however you feel, anyway, just an observation.

multihash_derive/src/lib.rs
```diff
@@ -0,0 +1,229 @@
+#![recursion_limit = "1024"]
```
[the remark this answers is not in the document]
dignifiedquire: tldr: it doesn't compile when I remove it

longer form:

as far as I understand the reason is that we are using macros, e.g. `quote!` inside the definition of another macro, and there is a recursion thing that breaks from the parsing level, but I am not entirely sure and haven't found any good docs on this yet

multihash_derive/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+                }
+
+                impl Code {
+                    pub fn from_u32(val: u32) -> Option<Self> {
+                        match val {
+                            #(#code_matches_from_u32)*
+                            _ => None
+                        }
+                    }
+                //     fn new_hasher(&self) -> Box<MultihashDigest> {
```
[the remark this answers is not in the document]
dignifiedquire: yes, but I want to revive it, but haven't figured out how yet

src/digests.rs
```diff
@@ -0,0 +1,213 @@
...
+
+impl AsRef<[u8]> for Multihash {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
+
+impl Multihash {
+    /// Creates a new `Multihash` from a `Vec<u8>`, consuming it.
+    /// If the input data is not a valid  multihash an error is returned.
```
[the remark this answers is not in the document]
dignifiedquire: :O

multihash_derive/src/lib.rs
dignifiedquire: Agreed.

multihash_derive/src/lib.rs
dignifiedquire: Size could be larger, (I probably should change it to a `u32`), and while it is currently restricted, I want to fix that later. By defining the signature here to be larger, I can add that later, without breaking any dependent code.

multihash_derive/src/lib.rs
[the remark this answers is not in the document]
dignifiedquire: this the best thing I found so far:

> Recursion limit

> The quote! macro relies on deep recursion so some large invocations may fail with "recursion limit reached" when you compile. If it fails, bump up the recursion limit by adding #![recursion_limit = "128"] to your crate. An even higher limit may be necessary for especially large invocations. You don't need this unless the compiler tells you that you need it.

multihash_derive/src/lib.rs
dignifiedquire: fixed to u23

multihash_derive/src/lib.rs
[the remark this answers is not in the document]
dignifiedquire: leaving this out for now

src/digests.rs
crackcomm: Can we get it to be `SHA2_256` etc.?

tomaka:
For what it's worth, we have a fork in the rust-libp2p repo that also contains a lot of changes: https://github.com/libp2p/rust-libp2p/tree/master/misc/multihash

(We did that in order to not have to wait for a review every time we needed something)

dignifiedquire:
@tomaka thanks for reminding me, I would like to make sure to get all things that are important for you back into here.  I‘ll review the commits there, but is there anything specific you have as a requirement you want to point out?

dignifiedquire:
it would also be great to get feedback if this branch is in a state that you could base your work on in rust-libp2p or if not what is missing.

dvc94ch:
interesting would be to add a cli app to compute hashes from stdin. would be useful to be able to do something like this: `cat file.txt | multihash --hash sha2-256 | multibase --base base32`

dvc94ch:
So is this getting merged?

mark-stopka:
Is this library still being maintained? SHA2 is already on a version 0.8.1; this PR also removes dependency on 0.7.x version of SHA2...

dvc94ch:
Yeah it's still being maintained, just currently busy with other things.

vmx:
I'm closing this PR as most of the features got implemented one way or another with the release of 0.10.0.

Missing is the `no_std` part, but I think that's something that can be tackled later in a separate PR.

PR #33, opened 2019-09-01
Add cli interface

dvc94ch:
As a well behaved unix library it should have a cli :)


dvc94ch:
Also adds murmur3 support for #34

dvc94ch:
@dignifiedquire what do you think?

dignifiedquire:
actually there was no fix, I just improved the api, so you could just use the original version here: https://github.com/stusmall/murmur3

vmx:
The API changed quite a bit with the release of 0.10.0 so this would need a rebase. If you still have interest in providing a CLI, please make this PR (or create a new one) with the CLI changes only and without adding the murmur hashing (that should be a separate PR).

dvc94ch:
I'll reopen, I think it would be useful to have

vmx:
I agree that it would be useful, thanks!

PR #42, opened 2020-01-17
multihash: Use `Bytes` instead of `Vec<u8>` internally. (#1187)

vmx:
multihash: Use `Bytes` instead of `Vec<u8>` internally.

To improve the efficiency of cloning multi-hashes (e.g. as the
representation of `PeerId`s), this PR replaces the `Vec<u8>`
representation with `Bytes`. The API is kept backwards
compatible and does not leak the representation type.

Originally from https://github.com/libp2p/rust-libp2p/commit/8af4a28152d12110dce72128bbe19a34106b1d00.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,8 @@ edition = "2018"
 [dependencies]
 blake2b_simd = { version = "0.5.9", default-features = false }
 blake2s_simd = { version = "0.5.9", default-features = false }
+bytes = "0.5"
 sha1 = "0.5"
 sha2 = { version = "0.7", default-features = false }
 tiny-keccak = "1.4"
-unsigned-varint = "0.2"
+unsigned-varint = "0.3"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,6 +13,7 @@ use std::convert::TryFrom;
 
 use blake2b_simd::blake2b;
 use blake2s_simd::blake2s;
+use bytes::{BufMut, Bytes, BytesMut};
 use sha2::Digest;
 use tiny_keccak::Keccak;
 use unsigned_varint::{decode, encode};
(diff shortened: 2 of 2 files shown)


dignifiedquire:
`Bytes` is being used less and less in the latest version of rust-libp2p, so I would be in favor of not doing this actually. 

@tomaka how important is this/how much does this actuall help in practice?

vmx:
@tomaka Just in case you wondering what's going on here. I currently work on getting all changes from rust-libp2p multihash back into this crate and then hopefully switch over rust-libp2p to this crate.

tomaka:
> @tomaka how important is this/how much does this actuall help in practice?

cc'ing @twittner

We have never actually benchmarked `Bytes` vs `Vec`.
`Bytes` is kind of the same of `Arc<[u8]>` while `Vec<u8>` is kind of the same as `Box<[u8]>`.
In situations where `Bytes` is beneficial, then it is very beneficial. In situations where it is not, the overhead should be very low.

twittner:
`Bytes` is usually not slower than `Vec<u8>` and whereas the latter's clone operation performance is dependent on data size, the former's is not. For multi-hash the data sizes involved are very small, so there is not much to gain. As for benchmarks, see https://gist.github.com/twittner/82dd2c83a3524b9ad2c84918183d008c.

vmx:
@twittner Would rust-libp2p use this crate (instead of the fork) if we keep using `Vec<u8>`? I think I'd prefer that as it keeps things simpler.

twittner:
My personal opinion is that I would prefer to use upstream, so if the only difference is `Vec<u8>` vs. `Bytes` it should not prevent rust-libp2p from adopting it I think. That being said I would also prefer it if you went with `Bytes` as using it internally has no downsides and it is better to avoid needless copies if possible.

vmx:
@twittner OK, then I misinterpreted your comment.

@dignifiedquire Are you OK going with `Bytes` then?

dignifiedquire:
not a fan, but I can live with it

dignifiedquire:
thank you @twittner and @tomaka for the details!

vmx:
I ran the benchmarks (thanks @twittner) locally and I'm seeing something like 10i-20% better perf on sizes 10-100 bytes (those are the ones that matter). It doesn't make the code much more complicated (well, except for adding a dependency) and our currently major user of this library prefers this change. Hence I decided to merge it.

PR #43, opened 2020-01-24
Add identity hash

vmx:
* Add peer id inlining for small public keys

* Apply @twittner suggestions

* Restore hashing

Originally from https://github.com/libp2p/rust-libp2p/commit/c6e8e6328ae15de27df64249be9085965ecf96fe.


vmx:
@dignifiedquire Added tests as CI passed.

PR #45, opened 2020-02-13
feat: Massive refactor with a new API

vmx:
The new API also makes it possible to wrap an existing hash into a Multihash. This
is useful if you e.g. want to create hashes for testing, without spending time with
actually hashing data.

You also interact with multicodecs less directly. This should make it easier to use
your own multihash implementations without forking the code base.

BREAKING CHANGE: There is a new API

When using multihashes, you now import implementations of that hash, which
has a `digest()` function. That function returns a Multihash.

New way:

    use multihash::{MultihashDigest, Sha3_512};
    let my_multihash = Sha3_512::digest(b"hello world!");

Old way:

    use multihash::{encode, Hash};
    let my_multihash = encode(Hash::SHA3512, b"hello world!");

This PR is heavily based on the ideas from https://github.com/multiformats/rust-multihash/pull/30, though I tried to keep things simple for now, not exposing a large API.

You might wonder why I implemented all those `MultihashDigest` implementations manually and didn't use a macro. The problem is that we are using different hashing libraries which don't have a unified API. Hence there are small differences that are also not easily expressible with macros. The [current version is using a macro](https://github.com/multiformats/rust-multihash/blob/d1214d5cbd4388e11de7d0d7f77a2160b5a67304/src/lib.rs#L25-L76) for that purpose, but I think that's neither easy to understand.

Though creating macros for specific algorithms makes sense, e.g. there could be one to support all Blake2 variants.

Please have a look a leave comments. This is really meant as a starting point for discussions, not as a "merge sooner rather than later" PR.


src/digests.rs
```diff
@@ -0,0 +1,211 @@
...
+    fn digest(&self, data: &[u8]) -> Multihash {
+        Self::digest(data)
+    }
+}
+
+/// Wraps a hash digest in Multihash with the given Mutlihash code.
+///
+/// The size of the hash is determoned by the size of the input hash. If it should be truncated
+/// the input data must already be the truncated hash.
+pub fn wrap(code: u64, data: &[u8]) -> Multihash {
```
dignifiedquire: This should probably take a `Code`, not a `u64`

vmx: The reason why this doesn't take a `Code` is, that it can be used for your custom, not yet implemented Multihash, without the need to fork the library. I thought about making it a trait, but that seemed a bit like overkill.

vmx:
I'm CC'ing a few folks that might have interest in reviewing this: @dvc94ch, @koivunej, @gozala

src/digests.rs
Gozala: @vmx I would suggest having a separate `fn` for that use case and have primary one be `Code` as that makes it more clear.

Gozala: Or better yet might be adding `Custom(u64)` to the enum so that use of non standard codes is explicit.

src/digests.rs
```diff
@@ -0,0 +1,211 @@
...
+impl Multihash {
+    /// Verifies whether `bytes` contains a valid multihash, and if so returns a `Multihash`.
+    pub fn from_bytes(bytes: Vec<u8>) -> Result<Multihash, DecodeOwnedError> {
+        if let Err(err) = MultihashRef::from_slice(&bytes) {
+            return Err(DecodeOwnedError {
+                error: err,
+                data: bytes,
+            });
+        }
+        Ok(Multihash {
```
Gozala: As far as I understand it `MultihashRef` is dropped here but we know `bytes` are valid or we'd have `Err`. I do however find it confusing / awkward, probably my lack of  Rust experience, makes me react with [Parse don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)

I think parse approach would not work here as `MultihashRef` contains  `&[u8]`, but maybe it make sense to have separate fn that does not construct `MultihashRef` ?

Maybe this is idiomatic in Rust, but maybe a comment for newbies (like myself) would be worth it.

src/digests.rs
```diff
@@ -0,0 +1,211 @@
...
+            bytes: Bytes::from(bytes),
+        })
+    }
+
+    /// Returns the bytes representation of the multihash.
```
Gozala: I apologize if I'm misunderstanding this but as far as I understand this would return bytes without consuming `Multihash` as `to_vec` creates a copy (isn't it). This seems to contradict expectation from [`into_bytes` in `String`s](https://doc.rust-lang.org/std/string/struct.String.html#method.into_bytes).

If my assumption is correct this method probably should either be removed or changed so it consumes multihash to match String behavior.

src/digests.rs
```diff
@@ -0,0 +1,211 @@
...
+            varint_decode::u64(&bytes).map_err(|_| DecodeError::BadInputLength)?;
+        if (bytes.len() as u64) != hash_len {
+            return Err(DecodeError::BadInputLength);
+        }
+
+        Ok(MultihashRef { bytes: input })
+    }
+
+    /// Returns which hashing algorithm is used in this multihash.
+    pub fn algorithm(&self) -> Code {
```
Gozala: So wrap allows creating a multihash with non standard code, but such multihash will cause panic if it's `algorithm()` is called ? I think it's better to decide whether custom algorithms are ok or not and stick to that decision than this.

Personally I would prefer to have custom `Code` support so it's supported but also clear when non-standard version is used.

src/digests.rs
```diff
@@ -0,0 +1,211 @@
...
+        let code = varint_decode::u64(&self.bytes)
+            .expect("multihash is known to be valid algorithm")
+            .0;
+        Code::from_u64(code).expect("multihash is known to be valid")
+    }
+
+    /// Returns the hashed data.
+    pub fn digest(&self) -> &'a [u8] {
```
Gozala: I would much rather have no way to creating invalid `Multihash` than having to deal with the possibly invalid multihash across it's methods.

src/digests.rs
rklaehn: I think this is correct. A method named `into_` carries with it the expectation that it consumes self and can thus perform the conversion in a cheap way.

For naming guidelines, see https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

src/digests.rs
```diff
@@ -0,0 +1,211 @@
...
+use num_traits::cast::FromPrimitive;
+use unsigned_varint::{decode as varint_decode, encode as varint_encode};
+
+use crate::errors::{DecodeError, DecodeOwnedError};
+use crate::hashes::Code;
+
+/// Represents a valid multihash.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Multihash {
+    bytes: Bytes,
```
rklaehn: Just curious: why are we using Bytes here? Bytes is good if you want to share multiple slices of some larger buffer without copying. But a multihash is typically self-contained. Using bytes here might have the advantage of avoiding a copy, but it carries the risk of dragging around much more memory than we actually need.

E.g. you get a large buffer from some network or disk interface as a Bytes, then take the first 32 bytes from the buffer into a multihash. This will then contain a pointer to the large buffer even if you keep the multihash for long and are no longer interested in the rest of the buffer.

Now, sharing is obviously important for a Multihash, but I would think that just using an Arc<[u8]> (not Arc<Vec<u8>> !) would be the best compromise.

E.g..

Gozala: Was also wondering about this but forgot to ask. Thanks @rklaehn for bringing it up

src/digests.rs
```diff
@@ -0,0 +1,211 @@
...
+    fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
+        Multihash::from_bytes(value)
+    }
+}
+
+/// Represents a valid multihash.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct MultihashRef<'a> {
+    bytes: &'a [u8],
+}
```
koivunej: Since this hasn't been brought up before, structures like this make me wonder if `Multihash` should be 

```
struct Multihash<'a> {
  bytes: std::borrow::Cow<'a, [u8]>,
}
```

This would remove the need for separate `Multihash` and `MultihashRef<'_>` being able to support both owning and working on borrowed data. This does bring some less ergonomic points in the longer run but as long as there is just some `bytes` this might be doable. Perhaps there was some rationale for this split which is not in any of the comments?

rklaehn: I think a multihash with a lifetime would be a world of hurt in terms of ergonomics. I can already see myself fighting lifetime issues to do utterly trivial things with multihashes and futures.

Also, I kinda dislike Cow since it has some surprising footguns... E.g. https://github.com/rust-lang/rust/issues/50160

I think an `Arc<[u8]>` would be best.

src/digests.rs
koivunej: I also wonder if dependency on `bytes` is really needed here. What was the original rationale (bytes vs. `Vec<_>`)?

For small amounts of data (like hashes) it might be always faster to just allocate a vector, like [sled might be doing in future](https://github.com/spacejam/sled/pull/942) than to do anything related to Arcs. Then again, perhaps my knowledge of `bytes` internals is outdated and I don't have time right now to understand them fully, could be that using `bytes` for single owner data is simple as `Vec<_>`.

Even if it was, as `bytes` is part of the async ecosystem it might be easier to get this crate stable throught the use of only mostly `std` components and to need to update this crate only when source specs or hash implementations evolve or more are needed.

Regarding @rklaehn from network usecase, I wonder if there is any benefit from actually constructing an owned type around the bytes, as the bytes are usable for verification purposes through the borrowed type, whatever it ends up looking like.

Also wondering if there should be a way to hash into a hopefully large enough `&mut [u8]` but perhaps that can be experimented outside of the crate if necessary. The same goes for updating on blocks. Perhaps this crate is at a quite difficult intersection with simplicity and different advanced IO usecases. I also wonder if only mapping from the spec codes to digest instances which can be used the by the caller in number of different ways would be enough.

src/digests.rs
koivunej: I agree, to use `Cow` can bring some interesting fights with borrowck. To use with owned data you could have `type OwnedMultihash = Multihash<'static>;` and make sure to always deal with that. I understand that current limitations with `async fn` push many things towards needing to be always owned or basically `Arc<_>` but at the same time I am not sure if a crate like `multihash` is a good place decide require anything more than the minimum (slices).

The downside of using `std::borrow::Cow` would bring the additional requirement of being able to use only `Vec` as the owned companion, so perhaps keeping the `MultihashRef<'_>` allows for more use cases.

src/digests.rs
rklaehn: For the case where you don't want to own the thing, there is `MultihashRef`. And in general I must say fewer dependencies are always better, especially for crates that have not yet reached 1.0.

tomaka: > Even if it was, as bytes is part of the async ecosystem

Just to react to that: the `bytes` crate was introduced as a hack around the fact that you couldn't do anything combining asynchronous and borrowing. You couldn't for instance asynchronously write a buffer by reference.

This is no longer true in a post-async-await world, and I totally see `bytes` disappearing at some point.

vmx: The background why `bytes` is used: https://github.com/multiformats/rust-multihash/pull/42.

Gozala: I would propose comment summarizing why `Bytes` are used, clearly without a good context it seems controversial.

vmx:
I've pushed a [new commit](https://github.com/multiformats/rust-multihash/pull/45/commits/bc3f9bb6f3e1c73eaf7ac9414bbd5733f9d0b640) addressing the use of `Code` instead of a raw integer. How do you like that approach @Gozala and @dignifiedquire?

I wasn't sure how usable that would be but looking at the test for the custom codec i quite like it (though if there's a better way, let me know).

src/hashes.rs
```diff
@@ -1,110 +1,305 @@
...
-            Hash::Keccak224 => 0x1A,
-            Hash::Keccak256 => 0x1B,
-            Hash::Keccak384 => 0x1C,
-            Hash::Keccak512 => 0x1D,
-            Hash::Blake2b256 => 0xB220,
-            Hash::Blake2b512 => 0xB240,
-            Hash::Blake2s128 => 0xB250,
-            Hash::Blake2s256 => 0xB260,
+    Blake2s256 = 0xb260,
+}
```
austinabell: There is no way to go from these codes to the structs used for generating digests? This also limits functionality in some dynamic cases where if you get the code from calling `algorithm(&self) -> Code` you couldn't generate another hash from this. Just a thought, could have a function which takes the code as input and returns a generic type bounded on `DynMultihashDigest` to solve this problem?

src/digests.rs
rklaehn: The discussion in https://github.com/multiformats/rust-multihash/pull/42 was comparing Bytes to Vec<u8>. I would love to see the tradeoff between Bytes and Arc<[u8]>. It might be that there are situations where Bytes comes up on top, but currently I don't see them.

src/hashes.rs
```diff
@@ -1,110 +1,305 @@
...
-            Hash::SHA3384 => 48,
-            Hash::SHA3512 => 64,
-            Hash::Keccak224 => 28,
-            Hash::Keccak256 => 32,
-            Hash::Keccak384 => 48,
-            Hash::Keccak512 => 64,
-            Hash::Blake2b256 => 32,
-            Hash::Blake2b512 => 64,
-            Hash::Blake2s128 => 16,
-            Hash::Blake2s256 => 32,
```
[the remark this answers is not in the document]
vmx: The reason for removing `size()` is, that it currently doesn't do what I would expect it to do. `size()` should return the size of the digest in the multihash. Though you can also use a truncated version of the hash. Hence there could e.g. be a Sha2-256 hash trunctated to 10 bytes. Then `size()` should obviously be `10` and not `32`.

In the link you provide it is using `digest.len()`. Shouldn't that still work as that's the size we really want?

src/hashes.rs
vmx: That's a great idea! See the latest commits for an implementation.

src/hashes.rs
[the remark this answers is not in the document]
austinabell: yeah there shouldn't really be a use for this, I was just indicating the multihash length was included in the prefix type you guys had created in the Cid library (but there was no actual need for that specific field) so if you wanted to get the length of the hash type you would actually have to hash some arbitrary data (again this isn't a problem as I don't see why you would need this)

src/hashes.rs
```diff
@@ -1,110 +1,339 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub enum Code {
+    /// Make it possible to use a custom code that is not part of the enum yet
+    Custom(u64),
+    /// Hashes that are immplemented in this crate
+    Implemented(ImplementedCode),
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum ImplementedCode {
```
austinabell: Thoughts on removing this type and just moving the implemented codes into the Code enum? Seems redundant and annoying to use a seperate type for this. Something like https://github.com/austinabell/rust-multihash/commit/da0d002d5ec3c3e5e64105ddee175f5d2839284c perhaps?

src/digests.rs
```diff
@@ -0,0 +1,206 @@
...
+
+impl<'a> PartialEq<Multihash> for MultihashRef<'a> {
+    fn eq(&self, other: &Multihash) -> bool {
+        self.bytes == &*other.bytes
+    }
+}
+
+/// The `MultihashDigest` trait specifies an interface common for all multihash functions.
+pub trait MultihashDigest {
```
austinabell: What is the functional benefit to having the builtin functionality tied to this trait? Seems unnecessary to have to import the MultihashDigest trait just to use a builtin hash. I only see the need for the DynMultihashDigest trait, unless I'm missing the use case for this trait?

src/digests.rs
vmx: I've updated the Gist (https://gist.github.com/vmx/0dc0d981b522e2834084559faa0dcb56) and added the code from https://github.com/multiformats/rust-multihash/pull/47 as comparison. On my machine that new code is not slower than the bytes approach.

src/hashes.rs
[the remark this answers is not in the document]
vmx: @austinabell Thanks for bringing this up. I agree I don't see why it would be a problem in practice.

src/hashes.rs
vmx: Good point and good idea. I wanted that, but you can't mix `Custom(u64)` together with definitions like `Sha2_512 = 0x13`. But what I haven't thought of was doing it your way. :)

src/digests.rs
vmx: I see this trait as the thing you would actually want. The `DynMultihashDigest` is a workaround around Rust limitations.

What do others think? Does it make sense to have both traits, or should we only have the hacky `DynMultihashDigest`?

austinabell: to be clear about what I am suggesting: https://github.com/austinabell/rust-multihash/commit/131cc9470f282967031da50c551170f001db5580 

Just quickly put together, could add code as another base impl as non associative and it would be functionally the same but not require you to import this trait whenever you want to hash in a non-dynamic context

vmx: Having `code()` as a method isn't a big deal. Though the problem I'm seeing is that now `code()` and `digest()` will be instance methods (right word?) that take `self` as first parameter and not static methods. To me, the API should be clean and not hide that `self` isn't actually needed. This is what I meant with `DynMultihashDigest` being hacky.

vmx:
This issue has already lots of comments, hence here's a summary of the current state (if I missed anything, let me know):

 Resolved:
  - `Bytes` vs. `Vec<u8> vs. something else: that discussion has been moved to https://github.com/multiformats/rust-multihash/pull/47
  - Using `Code` instead of a raw int for wrapping a Multihash
  - Agreement that this new API is better than the old one

Open:
 - Having the `DynMultihashDigest` trait *and* the `MultihashDigest` or just the `DynMultihashDigest` one (https://github.com/multiformats/rust-multihash/pull/45#discussion_r380265109)

src/digests.rs
austinabell: Yeah I was just too lazy to move the associated code const to the base impl, does this satisfy that criteria: https://github.com/austinabell/rust-multihash/commit/b8e5c0278648e56d012ce17bcc9893ba34ae9dee ?

austinabell: Just to be clear, I have no opinion or preference, just offering an alternative since it seems weird to me to require importing that trait that does nothing just to be able to hash using builtin implementations. Seems to be an even tradeoff since declaring custom implementations could be slightly confusing

vmx: That's (from the end-user perspective) quite nice. We could also have a derive macro as the trait implementation would be the same.

Could please someone with more Rust experience comment on which approach is better? The one proposed by @austinabell (https://github.com/multiformats/rust-multihash/pull/45#discussion_r380332594) feels a bit un-idiomatic, but the resulting API is quite nice. @dignifiedquire perhaps?

vmx:
I like the idea from @austinabell and applied his commit. The `MutlihashDigest` trait implementations can easily be replaced with a macro. But I think that could be done later on (even on a separate PR).

This resolves all open issues.

@dignifiedquire I'm happy with this PR now, it's ready for a final review.

src/digests.rs
spacejam: re: sled use, I'm closing that linked issue because although it is faster to copy 2-4k than bump an Arc in some cases, it's NOT faster to allocate a vec first. I need to do more measuring to find out whether that third path between inlining and Arc<[u8]> actually exists.

vmx:
I rebased it (which was more work than expected) on master. It would be cool to get another quick review by someone to make sure I didn't make any major mistake during this rebase.

vmx:
I couldn't resist and pushed one more commit into this PR as passing on the code by reference felt awkward.

Issue #46, opened 2020-02-15
Explore using inline data storage


rklaehn:
The most common hashes are 32 bytes long. So with the additional data that gives 34 bytes. It seems inefficient to store such a small object on the heap with the associated overhead / pointer chasing / cache issues.

It would be possible to use an approach similar to https://crates.io/crates/smallvec and store hashes inline up to a certain size, and only store on the heap once the needed size is bigger.

Note that we would not want to use smallvec since we don't need all the metadata of a vec. The hash already knows how big it is, so what we would need would be just an enum like this:

```rust
enum Storage {
  Inline([u8; 39]), // any hash up to 39 bytes goes in here
  Heap(Arc<[u8]>), // the rare hash that is larger goes in here
}
```

This can be made pretty much opaque from the outside, like smallvec. Might not even need unsafe...

rklaehn:
Overall size would be 40 bytes, still smaller than the typical cache line size.

vmx:
That improvement got merged.

PR #47, opened 2020-02-15
Use inline storage for small hashes

rklaehn:
This also gets rid of the Bytes dependency and replaces it with an `Arc<[u8]>`.

Implementation for https://github.com/multiformats/rust-multihash/issues/46

Note that the size of this new multihash is 40, where as the size of the old Bytes based multihash is 32, plus the stuff on the heap. I think this is a pretty good tradeoff!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -161,7 +161,7 @@ pub struct Multihash {
 
 impl Debug for Multihash {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "Multihash")
+        f.debug_tuple("Multihash").field(&self.as_bytes()).finish()
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,18 +8,22 @@
 
 mod errors;
 mod hashes;
+mod storage;
 
 use std::convert::TryFrom;
+use std::fmt::Debug;
+use std::hash;
 
 use blake2b_simd::{blake2b, Params as Blake2bVariable};
 use blake2s_simd::{blake2s, Params as Blake2sVariable};
-use bytes::{BufMut, Bytes, BytesMut};
 use sha2::Digest;
 use tiny_keccak::Keccak;
 use unsigned_varint::{decode, encode};
 
 pub use errors::{DecodeError, DecodeOwnedError, EncodeError};
 pub use hashes::Hash;
+use std::fmt;
+use storage::Storage;
 
 // Helper macro for encoding input into output using sha1, sha2, tiny_keccak, or blake2
 macro_rules! encode {
@@ -107,12 +111,12 @@ pub fn encode(hash: Hash, input: &[u8]) -> Result<Multihash, EncodeError> {
 
         let total_len = code.len() + size.len() + input.len();
 
-        let mut output = BytesMut::with_capacity(total_len);
-        output.put_slice(code);
-        output.put_slice(size);
-        output.put_slice(input);
+        let mut output = Vec::with_capacity(total_len);
+        output.extend_from_slice(code);
+        output.extend_from_slice(size);
+        output.extend_from_slice(input);
         Ok(Multihash {
-            bytes: output.freeze(),
+            storage: Storage::copy_from_slice(&output),
         })
     } else {
         let (offset, mut output) = encode_hash(hash);
@@ -135,31 +139,51 @@ pub fn encode(hash: Hash, input: &[u8]) -> Result<Multihash, EncodeError> {
         });
 
         Ok(Multihash {
-            bytes: output.freeze(),
+            storage: Storage::copy_from_slice(&output),
         })
     }
 }
 
-// Encode the given [`Hash`] value and ensure the returned [`BytesMut`]
+// Encode the given [`Hash`] value and ensure the returned [`Vec<u8>`]
 // has enough capacity to hold the actual digest.
-fn encode_hash(hash: Hash) -> (usize, BytesMut) {
+fn encode_hash(hash: Hash) -> (usize, Vec<u8>) {
     let mut buf = encode::u16_buffer();
     let code = encode::u16(hash.code(), &mut buf);
 
     let len = code.len() + 1 + usize::from(hash.size());
 
-    let mut output = BytesMut::with_capacity(len);
-    output.put_slice(code);
-    output.put_u8(hash.size());
+    let mut output = Vec::with_capacity(len);
+    output.extend_from_slice(code);
+    output.push(hash.size());
     output.resize(len, 0);
 
     (code.len() + 1, output)
 }
 
 /// Represents a valid multihash.
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Clone)]
 pub struct Multihash {
-    bytes: Bytes,
+    storage: Storage,
+}
+
+impl Debug for Multihash {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Multihash")
+    }
+}
+
+impl PartialEq for Multihash {
+    fn eq(&self, other: &Self) -> bool {
+        self.storage.bytes() == other.storage.bytes()
+    }
+}
+
+impl Eq for Multihash {}
+
+impl hash::Hash for Multihash {
+    fn hash<H: hash::Hasher>(&self, state: &mut H) {
+        self.storage.bytes().hash(state);
+    }
 }
 
 impl Multihash {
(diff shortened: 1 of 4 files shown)


src/storage.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// The raw bytes. Note that this can be longer than the data this storage has been created from.
+    pub fn bytes(&self) -> &[u8] {
+        match self {
+            Storage::Inline(bytes) => bytes,
+            Storage::Heap(data) => &data,
+        }
+    }
+
+    /// creates storage from a vec. Note that this will not preserve the size.
+    pub fn copy_from_slice(slice: &[u8]) -> Self {
```
burdges: I think the `copy_from_slice` overwrites existing slices, so you want `from_slice` here

src/storage.rs
```diff
@@ -0,0 +1,43 @@
+use std::sync::Arc;
+
+const MAX_INLINE: usize = 39;
```
dignifiedquire: why 39?

rklaehn: So the total size is 40, a multiple of 8. 1 byte is needed for the enum discriminator.

rklaehn: 39 fits all 256 bit hashes with some room to spare. 512 byte hashes won't work though, and I think once you go to such large hashes, you are better off with an `Arc<[u8]>`.

But the size of the inline buffer can of course be adjusted when even larger hashes become common. The whole mechanism is completely opaque.

rklaehn:
@dignifiedquire smallvec has an overhead of one usize for storing the size https://github.com/servo/rust-smallvec/blob/master/lib.rs#L374 , which is not necessary because the hash already knows how big it is. It also has the overhead of the discriminator, unless you use the "union" feature. https://github.com/servo/rust-smallvec/blob/master/lib.rs#L292, which is disabled by default: https://docs.rs/smallvec/1.2.0/smallvec/#union-feature . Due to the alignment rules, this discriminator costs you 8 bytes on 64 bit arch. So in total an overhead of 16 bytes.

But most of all, it is a dependency containing unsafe code that can cause problems for others and that is not needed for a few LOC.

burdges:
There is an old [smallbox](https://github.com/andylokandy/smallbox) crate but comes with a silly [inefficiency](https://andylokandy.github.io/smallbox/doc/src/smallbox/smallbox.rs.html#67-71) that's worse than smallvec here.

I'd think fix `SmallBox<T, Space>` with an inner union that discriminates based on `size_of::<T>() < size_of::<Space>()`, except..

I presume smallbox really exists to make `SmallBox<dyn Trait, Space>` work.  Ideally, you'd make two vtables for `dyn Trait` that distinguished between smaller and larger than `Space`, but this would require deep rustc integration. 

It appears possible to do this manually for individual traits, like we really could make a no_std no_alloc compatible `SmallError` that only takes 16 bytes and lets `std::io` traits work without std or alloc, so long as error types stay smaller than `usize`: https://github.com/rust-lang/rfcs/pull/2820#issuecomment-567439210

It's possible some `#[dyn_smallbox]` procmaco could abstract this over traits with the current rustc, but nobody did one.

Apologies for the derail, but I'm agreeing with you that smallbox optimizations become messy and custom, especially whenever you want `dyn Trait` to work.

rklaehn:
@burdges interesting, but I think in this case just doing the branching manually is best. It is quite simple code.

rklaehn:
Thought about this some more.
- Given that the average 256 bit hash is 32 bytes, 34 bytes with metadata, there is enough room to store the length separately in 40 bytes to avoid this awkward recomputation of the length from the content. Will update the PR. Then the storage is just an efficient encoding of an arbitrary Vec<u8>, giving better separation of concerns.
- The way Bytes was used here, there is really no benefit in Bytes over Arc<[u8]>.

src/storage.rs
```diff
@@ -0,0 +1,55 @@
+use std::sync::Arc;
+
+const MAX_INLINE: usize = 38;
```
vmx: Could you please add a comment here, why it is 38.

rklaehn:
@vmx thanks for the benchmark. Criterion is pretty nice.

For small hashes that fit, the improvement is quite good. I think we should give multiaddr the same treatment once this is merged...

vmx:
@twittner Could you please have a look at this PR? You were in favour for using `Bytes`. Would this change for for your use case in libp2p?

src/lib.rs
```diff
@@ -107,12 +111,12 @@ pub fn encode(hash: Hash, input: &[u8]) -> Result<Multihash, EncodeError> {
...
-        output.put_slice(code);
-        output.put_slice(size);
-        output.put_slice(input);
+        let mut output = Vec::with_capacity(total_len);
+        output.extend_from_slice(code);
+        output.extend_from_slice(size);
+        output.extend_from_slice(input);
         Ok(Multihash {
-            bytes: output.freeze(),
+            storage: Storage::from_slice(&output),
```
twittner: Instead of always copying from a temporary `Vec`, how about `Storage` gets a `Storage::from_slices` method and decides internally if the accumulated length fits into its inline storage in which case we avoid the heap allocation altogether, e.g.

```rust
    pub(crate) fn from_slices(slices: &[&[u8]]) -> Self {
        let n = slices.iter().fold(0usize, |a, s| a.saturating_add(s.len()));
        if n <= MAX_INLINE {
            let s = slices.iter().fold(([0; MAX_INLINE], 0), |(mut array, i), s| {
                array[i .. i + s.len()].copy_from_slice(s);
                (array, i + s.len())
            });
            Storage::Inline(n as u8, s.0)
        } else {
            let mut v = Vec::with_capacity(n);
            for s in slices {
                v.extend_from_slice(s)
            }
            Storage::Heap(v.into())
        }
    }
```

src/lib.rs
```diff
@@ -135,31 +139,51 @@ pub fn encode(hash: Hash, input: &[u8]) -> Result<Multihash, EncodeError> {
         });
 
         Ok(Multihash {
-            bytes: output.freeze(),
+            storage: Storage::from_slice(&output),
```
twittner: Since `output` is always a `Vec`, how about `Storage::Heap(output.into())`?

src/lib.rs
```diff
@@ -172,7 +196,7 @@ impl Multihash {
             });
         }
         Ok(Multihash {
-            bytes: Bytes::from(bytes),
+            storage: Storage::from_slice(&bytes),
```
twittner: Like above, why not keep the existing `Vec` and use `Storage::Heap(bytes.into())`?

src/storage.rs
```diff
@@ -0,0 +1,61 @@
...
+/// MAX_INLINE is the maximum size of a multihash that can be stored inline
+///
+/// We want the currently most common multihashes using 256bit hashes to be stored inline. These
+/// hashes are 34 bytes long. An overall size of 38 seems like a good compromise. It allows storing
+/// any 256bit hash with some room to spare and gives an overall size for Storage of 40 bytes, which
+/// is a multiple of 8. We need 2 extra bytes, one for the size and one for the enum discriminator.
+const MAX_INLINE: usize = 38;
+
+#[derive(Clone)]
+pub enum Storage {
```
twittner: Personally I would still use `pub(crate) enum Storage` here and in the methods of this type even though the mod is private, but it is only a minor detail and I guess a matter of preference.

src/lib.rs
rklaehn: Yes, dealing with this temporary Vec has been on my mind, but I thought it would be best to keep it for the next PR, since it is a separate concern.

src/lib.rs
rklaehn: If the vec contains <= MAX_INLINE bytes, you don't want Heap.

For the case where inline storage is not possible, do you think that .into() is more efficient to go from a `Vec<u8>` to an `Arc<[u8]>`?

rklaehn: Looking at the From instance that does the conversion, going from a Vec<u8> to a Arc<[u8]> does not seem to gain you anything compared to just copying from a slice.

https://doc.rust-lang.org/src/alloc/sync.rs.html#2021-2033

src/lib.rs
rklaehn: This is only for the identity hash. Will add this and add a test.

src/lib.rs
```diff
@@ -25,6 +25,10 @@ pub use hashes::Hash;
 use std::fmt;
 use storage::Storage;
 
+#[cfg(test)]
+#[macro_use]
+extern crate quickcheck;
```
vmx: Could you please remove those and instead…

src/storage.rs
```diff
@@ -37,11 +37,32 @@ impl Storage {
...
+            }
+            Storage::Heap(v.into())
+        }
+    }
 }
 
 #[cfg(test)]
 mod tests {
     use super::{Storage, MAX_INLINE};
+    use quickcheck;
```
vmx: …import the macro via `use quickcheck::quickcheck`?

PR #50, opened 2020-02-20
Add ordering for Multihash and MultihashRef

rklaehn:
This is just lexicographical byte ordering, which should be very fast
in case of multihashes since they usually differ within the first 8 byte
word.

Ordering is useful to have if you e.g. want to output a set of multihashes
in a deterministic way for debugging/testing purposes. Also, many possible storage engines work using lexicographical byte ordering. And it allows you to use ordering based collections like BTreeSet with multihashes.

This is based on https://github.com/multiformats/rust-multihash/pull/47 , so should be reviewed once
that is merged.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -248,7 +248,13 @@ impl TryFrom<Vec<u8>> for Multihash {
 
 impl PartialOrd for Multihash {
     fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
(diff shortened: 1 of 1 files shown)


vmx:
@rklaehn Could you please rebase this change?

rklaehn:
@vmx Done. I also let rust derive the PartialOrd for the MultihashRef, since it is perfectly capable of figuring that out on its own...

Issue #48, opened 2020-02-19
Provide quickcheck arbitraries


rklaehn:
This would have to be behind an optional and off-by-default feature flag so you don't pay the price of the quickcheck dependency if you do not need it.

But arbitraries would be quite helpful for property based testing of more complex structures that use multihashes.

rklaehn:
Here is a little experiment how to add quickcheck arbitraries to a crate under a feature flag:
https://github.com/rklaehn/rust-feature-experiment

PR #51, opened 2020-02-20
Hash and Multihash arbitrary

rklaehn:
Implements #48 

This is a request for discussion about if this way of providing arbitraries both for internal use and for use from other crates is a good way to go.

If somebody has a simpler way to achieve this I am eager to hear about it...

I verified that the dependencies are not in the output unless you build with the feature flag enabled. See https://github.com/rklaehn/rust-feature-experiment

This is based on https://github.com/multiformats/rust-multihash/pull/47 , so should be reviewed once that is merged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,11 +18,16 @@ sha-1 = { version = "0.8", default-features = false }
 sha2 = { version = "0.8", default-features = false }
 sha3 = { version = "0.8", default-features = false }
 unsigned-varint = "0.3"
+quickcheck = { version = "0.9.2", optional = true }
+rand = { version = "0.7.3", optional = true }
 
 [dev-dependencies]
 criterion = "0.3"
-rand = "0.7"
 quickcheck = "0.9.2"
+rand = "0.7.3"
+
+[features]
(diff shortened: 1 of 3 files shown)


rklaehn:
Note that the tests that I have added in this crate using the arbitrary are in the files themselves.

The tests directory is for integration tests that use the crate externally, so there is not an easy way to enable the test feature for them, except maybe doing it on a file by file basis as outlined here: https://github.com/rust-lang/cargo/issues/2911#issuecomment-401952525

Cargo.toml
```diff
@@ -25,3 +25,12 @@ sha1 = "0.5"
 sha2 = { version = "0.7", default-features = false }
 tiny-keccak = "1.4"
 unsigned-varint = "0.3"
```
rklaehn: We need the dependencies twice, once for our own tests, and once for the test feature that other crates can use from their dev-dependencies.

rklaehn:
@vmx brought this up to date. WDYT? I would love the entire multiformats crate hierarchy to have arbs so that more complex crates such as rust-ipfs or rust-libp2p can use them to write property based tests / fuzz tests.

PR #58, opened 2020-03-07
WIP streaming support, macro refactor

sunny-g:
This PR does the following:

- extends `MultihashDigest` with an `input` and `result` method (borrowed from the `Digest` and `Input` traits) to support streaming
- implements the new methods for all existing codecs
  - this required changing their implementation to embed the existing digest, but the `impl Default` makes hasher construction easy
- drastically reduces the amount of repetition in the `impl Code` and `impl MultihashDigest`s using two local macros.
- replaces `Code::to_u64` and `Code::from_u64` with `From<Code> for u64` and `From<u64> for Code`

All the original tests still pass and the core API is mostly the same (with the exception of hasher construction) ideally satisfying the O in SOLID :grin: 

Let me know what y'all think!

### DIFF
--- a/src/digests.rs
+++ b/src/digests.rs
@@ -213,7 +213,7 @@ impl<'a> MultihashRef<'a> {
     pub fn algorithm(&self) -> Code {
         let (code, _bytes) =
             varint_decode::u64(&self.bytes).expect("multihash is known to be valid algorithm");
-        Code::from_u64(code)
+        Code::from(code)
     }
 
     /// Returns the hash digest.
@@ -313,7 +313,7 @@ pub trait MultihashDigest {
 /// ```
 pub fn wrap(code: Code, data: &[u8]) -> Multihash {
     let mut code_buf = varint_encode::u64_buffer();
-    let code = varint_encode::u64(code.to_u64(), &mut code_buf);
+    let code = varint_encode::u64(code.into(), &mut code_buf);
 
     let mut size_buf = varint_encode::u64_buffer();
     let size = varint_encode::u64(data.len() as u64, &mut size_buf);
--- a/src/digests.rs
+++ b/src/digests.rs
@@ -282,6 +282,18 @@ pub trait MultihashDigest {
     ///
     /// Panics if the digest length is bigger than 2^32. This only happens for identity hasing.
     fn digest(&self, data: &[u8]) -> Multihash;
+
+    /// Digest input data.
+    ///
+    /// This method can be called repeatedly for use with streaming messages.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the digest length is bigger than 2^32. This only happens for identity hashing.
+    fn input(&mut self, data: &[u8]);
+
+    /// Retrieve the computed `Multihash`, consuming the hasher.
+    fn result(self) -> Multihash;
 }
 
 /// Wraps a hash digest in Multihash with the given Mutlihash code.
--- a/src/digests.rs
+++ b/src/digests.rs
@@ -294,6 +294,14 @@ pub trait MultihashDigest {
 
     /// Retrieve the computed `Multihash`, consuming the hasher.
     fn result(self) -> Multihash;
+
+    /// Retrieve result and reset hasher instance.
+    ///
+    /// This method sometimes can be more efficient compared to hasher re-creation.
+    fn result_reset(&mut self) -> Multihash;
+
+    /// Reset hasher instance to its initial state.
+    fn reset(&mut self);
 }
 
 /// Wraps a hash digest in Multihash with the given Mutlihash code.
--- a/src/hashes.rs
+++ b/src/hashes.rs
@@ -238,49 +238,49 @@ impl Identity {
 derive_digest! {
     /// The SHA-1 hasher.
     @sha ::sha1::Sha1 as Sha1;
-        @code_doc "The code of the SHA-1 hasher, 0x11",
+        @code_doc "The code of the SHA-1 hasher, 0x11.",
     /// The SHA2-256 hasher.
     @sha ::sha2::Sha256 as Sha2_256;
-        @code_doc "The code of the SHA2-256 hasher, 0x12",
+        @code_doc "The code of the SHA2-256 hasher, 0x12.",
     /// The SHA2-512 hasher.
     @sha ::sha2::Sha512 as Sha2_512;
-        @code_doc "The code of the SHA2-512 hasher, 0x13",
+        @code_doc "The code of the SHA2-512 hasher, 0x13.",
     /// The SHA3-224 hasher.
     @sha ::sha3::Sha3_224 as Sha3_224;
-        @code_doc "The code of the SHA3-224 hasher, 0x17",
+        @code_doc "The code of the SHA3-224 hasher, 0x17.",
     /// The SHA3-256 hasher.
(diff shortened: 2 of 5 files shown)


src/digests.rs
```diff
@@ -282,6 +282,18 @@ pub trait MultihashDigest {
...
+    ///
+    /// This method can be called repeatedly for use with streaming messages.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the digest length is bigger than 2^32. This only happens for identity hashing.
+    fn input(&mut self, data: &[u8]);
+
+    /// Retrieve the computed `Multihash`, consuming the hasher.
+    fn result(self) -> Multihash;
```
dignifiedquire: should have result_reset and reset to allow easy reuse of the same instance

sunny-g: Thought it would pollute the interface and take away from the original intention of the Digest structs, but np!

sunny-g: Update: to do this, I think we'd need to replace  `blake2s_simd` with [`blake2`](https://crates.io/crates/blake2) as the current crate doesn't `impl Digest` or provide methods for resetting the hasher instance. `blake2` is written by the RustCrypto folks and also provides optional SIMD support, if required.

dignifiedquire: We should not switch, the speed difference is pretty large. We can implement resetting manually for crates that don’t support it

src/hashes.rs
```diff
@@ -1,473 +1,313 @@
...
+                fn result_reset(&mut self) -> Multihash {
+                    let digest = self.0.finalize();
+                    let hash = wrap(Self::CODE, digest.as_bytes());
+                    self.reset();
+                    hash
+                }
+                #[inline]
+                fn reset(&mut self) {
+                    self.0 = <$params>::new().hash_length($len).to_state();
+                }
```
sunny-g: Ok, my latest commit adds `result_reset` and `reset`, which for the Blake digests, just replaces the inner `blake2#_simd` digest with a new one.

src/hashes.rs
```diff
@@ -1,473 +1,313 @@
...
-}
-impl Sha2_256 {
-    /// The code of Sha2-256 hasher, 0x12.
-    pub const CODE: Code = Code::Sha2_256;
-    /// Hash some input and return the sha2-256 digest.
-    pub fn digest(data: &[u8]) -> Multihash {
-        let digest = sha2::Sha256::digest(&data);
-        wrap(Self::CODE, &digest)
+    fn reset(&mut self) {
+        self.0 = Vec::new();
```
vmx: Should this be `self.0.clear()` instead? I'm unsure what exactly the `reset()` semantics are supposed to be.

src/hashes.rs
```diff
@@ -1,473 +1,313 @@
...
-    /// Hash some input and return the blake2s-256 digest.
-    pub fn digest(data: &[u8]) -> Multihash {
-        let digest = Blake2s::new()
-            .hash_length(32)
-            .to_state()
-            .update(&data)
-            .finalize();
-        wrap(Self::CODE, &digest.as_bytes())
-    }
+derive_digest! {
```
vmx: I'm not sure about it myself, but would it make sense so split `derive_digest` into `derive_digest_sha` and `derive_digest_blake`? They don't seem to share code and they use different parameters. This might make the code easier to understand. What do you think?

src/hashes.rs
sunny-g: I think that's better, as the method docs only detail that it needs to "reset the hasher to its initial state": [docs](https://docs.rs/digest/0.8.1/digest/trait.Digest.html#tymethod.reset). I'll update when I get the chance.

src/hashes.rs
sunny-g: It'll ultimately look and behave the same - the only change would be swapping `@sha` with `sha` in the name of the macro.

src/hashes.rs
sunny-g: On second thought, the docs state that `clear()` "has no effect on the allocated capacity of the vector," which may or may not be desired behaviour as the `Vec` wouldn't literally be in its initial state...

vmx: I would say it would be in it's initial state, it's just not the same state.

I think the point of `reset()` is that it is an operation that is cheaper than recreating it. Hence `clear()` would make sense. You could always create a new object if you want to start from a clean state.

src/hashes.rs
vmx: If you care only about one of the macros while reading the source, it would be less cognitive load. Anyway, it's a matter of taste and there isn't a right or wrong. I'm happy to keep it as it currently is.

I'm using code review also as a chance to learn coding taste/decisions from others, hence I thought I also share mine :)

vmx:
Clippy (`cargo clippy --all-targets --all-features -- -D warnings`) is complaining about the complexity of some test functions. Could you please see if they could be made simpler or simply silence the warning for those functions.

src/hashes.rs
sunny-g: I appreciate the flexibility :)

src/hashes.rs
sunny-g: It's changed in the latest updates.

sunny-g:
My latest commit fixes the `cargo clippy` errors locally, don't know what's going on with the CI.

src/digests.rs
```diff
@@ -282,6 +282,26 @@ pub trait MultihashDigest {
     ///
     /// Panics if the digest length is bigger than 2^32. This only happens for identity hasing.
     fn digest(&self, data: &[u8]) -> Multihash;
```
dignifiedquire: One thing I like about the original `Digest` trait is that `digest` actually does not take `self`, but rather only the data, such that you can write `Sha2_256::digest([1u8; 14])` without going through creating an instance. I wonder if we should do this switch as well, to make the pi more ergonomic.

vmx: Isn't that still the case, with this code: https://github.com/multiformats/rust-multihash/pull/58/files#diff-5b11d6021f0af96d8da7e3de8ca657b4 Line 77-80 (sorry I can't directly link due to GitHub not showing the diff by default.

vmx: I re-read your comment, I think I now got the point. The reason to use `self` is, that it can be used as a trait object. I see the `MutlihashDigest` main use as a trait object, else you would use the impl directly, without the trait.

dignifiedquire: I guess, but then I think having the `fn digest(&self)` version makes even less sense

vmx: So `MultihashDigest` is kind of what https://docs.rs/digest/0.8.1/digest/trait.DynDigest.html is.

vmx: The idea is that you can call `digest()` based on a multihash (or just the code of it) without the need to know which algorithm it actually is: `my_multihash.algorithm().hasher()?.digest("something")`.

dignifiedquire: I see, that works for me, but should probably be documented somewhere :)

vmx:
Thanks a lot @sunny-g for your work. I won't do a new release immediately as I'd like to do another minor release as this might break people's code. I've more changes I'd like to get into that release, e.g. https://github.com/multiformats/rust-multihash/pull/60 (which needs to be updated to work with the current code).

sunny-g:
@vmx hey, no problem! Glad to see this get resolved and I appreciate everyone's feedback!

PR #60, opened 2020-03-17
 feat: use generics for custom code tables

vmx:
With generics it is now possible to have your own code tables. Most users
won't need those, hence there are type aliases to make this library easy
to use with the default case.

This doesn't commit doesn't change the API of `Multihash` or `MultihashRef`,
but the `MultihashDigest` trait now needs a generic type, usually
`MultihashDigest<Code>`.

BREAKING CHANGE: `MultihashDigest` needs a generic now

If you have used `MultihashDigest` you now need to use it as `MultihashDigest<Code>`.

### DIFF
--- a/src/digests.rs
+++ b/src/digests.rs
@@ -131,7 +131,7 @@ pub type MultihashRef<'a> = MultihashRefGeneric<'a, Code>;
 /// );
 /// ```
 #[derive(Clone)]
-pub struct MultihashGeneric<T> {
+pub struct MultihashGeneric<T: TryFrom<u64>> {
     storage: Storage,
     // Use `PhantomData` in order to be able to make the `Multihash` struct take a generic
     _code: PhantomData<T>,
@@ -143,15 +143,15 @@ impl<T: TryFrom<u64>> fmt::Debug for MultihashGeneric<T> {
     }
 }
 
-impl<T> PartialEq for MultihashGeneric<T> {
+impl<T: TryFrom<u64>> PartialEq for MultihashGeneric<T> {
     fn eq(&self, other: &Self) -> bool {
         self.storage.bytes() == other.storage.bytes()
     }
 }
 
-impl<T> Eq for MultihashGeneric<T> {}
+impl<T: TryFrom<u64>> Eq for MultihashGeneric<T> {}
 
-impl<T> hash::Hash for MultihashGeneric<T> {
+impl<T: TryFrom<u64>> hash::Hash for MultihashGeneric<T> {
     fn hash<H: hash::Hasher>(&self, state: &mut H) {
         self.storage.bytes().hash(state);
     }
@@ -407,7 +407,7 @@ impl<'a, T: TryFrom<u64>> Into<Vec<u8>> for MultihashRefGeneric<'a, T> {
 }
 
 /// The `MultihashDigest` trait specifies an interface common for all multihash functions.
-pub trait MultihashDigest<T> {
+pub trait MultihashDigest<T: TryFrom<u64>> {
     /// The Mutlihash byte value.
     fn code(&self) -> T;
 
@@ -454,7 +454,7 @@ pub trait MultihashDigest<T> {
 /// let wrapped: Multihash = wrap(Code::Sha2_256, &digest);
 /// assert_eq!(wrapped.digest(), digest);
 /// ```
-pub fn wrap<T: Into<u64> + std::fmt::Debug>(code: T, data: &[u8]) -> MultihashGeneric<T> {
+pub fn wrap<T: Into<u64> + TryFrom<u64>>(code: T, data: &[u8]) -> MultihashGeneric<T> {
     let mut code_buf = varint_encode::u64_buffer();
     let code = varint_encode::u64(code.into(), &mut code_buf);
 
--- a/src/digests.rs
+++ b/src/digests.rs
@@ -25,7 +25,7 @@ pub type MultihashRef<'a> = MultihashRefGeneric<'a, Code>;
 /// # Example
 ///
 /// ```
-/// use multihash::{wrap, MultihashDigest, MultihashGeneric};
+/// use multihash::{wrap, MultihashGeneric};
 /// use std::convert::TryFrom;
 ///
 /// #[derive(Debug)]
--- a/src/digests.rs
+++ b/src/digests.rs
@@ -222,7 +222,7 @@ impl<T: TryFrom<u64>> MultihashGeneric<T> {
     /// ```
     pub fn algorithm(&self) -> T
     where
-        <T as std::convert::TryFrom<u64>>::Error: std::fmt::Debug,
+        <T as TryFrom<u64>>::Error: std::fmt::Debug,
     {
         self.as_ref().algorithm()
     }
@@ -343,11 +343,11 @@ impl<'a, T: TryFrom<u64>> MultihashRefGeneric<'a, T> {
     /// ```
     pub fn algorithm(&self) -> T
     where
-        <T as std::convert::TryFrom<u64>>::Error: std::fmt::Debug,
+        <T as TryFrom<u64>>::Error: std::fmt::Debug,
     {
         let (rawcode, _bytes) =
             varint_decode::u64(&self.bytes).expect("multihash is known to be valid algorithm");
-        T::try_from(rawcode).unwrap()
+        T::try_from(rawcode).expect("multihash is known to be a valid algorithm")
     }
 
     /// Returns the hash digest.
--- a/src/digests.rs
+++ b/src/digests.rs
@@ -1,5 +1,6 @@
 use std::borrow::Borrow;
-use std::convert::TryFrom;
+use std::convert::{Into, TryFrom};
+use std::marker::PhantomData;
 use std::{cmp, fmt, hash, ops};
 
 use unsigned_varint::{decode as varint_decode, encode as varint_encode};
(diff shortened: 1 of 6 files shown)


vmx:
Before I publish a rebased version (I'd expect the changes to be harder to follow), I'd like to get feedback if you think this change makes sense. This PR has to be seen in a wider context together with rust-cid and rust-ipld as there you might also want to use your own custom codec tables.

/cc @dvc94ch @rklaehn

vmx:
Also cc @gozala.

dvc94ch:
mmh. I see what you are doing. So the idea would be to create an `enum CustomCode` and parameterize the digest over `CustomCode`. The problem with this is that they might be different codes refering to the same algorithm. Which could be a good thing if someone wants to use a different implementation. But overall it adds complexity. Can you maybe add an example of how one would add a custom hash now vs how one would have done it before?

vmx:
> Can you maybe add an example of how one would add a custom hash now vs how one would have done it before?

Adding a custom in this code is here: https://github.com/multiformats/rust-multihash/pull/60/files#diff-cd318588a0f874eb790d208a26a86ab0R70-R132

Before this change we had `Custom(u64)` within the enum. You also couldn't really add a custom hasher to it transparently, you would need to match whether it is a custom code and then return the hasher in your application logic.

I would argue that custom hashes are rare, hence I'm happy to take the trade-off of being it a bit harder to do, but still possible.

One use case for a custom code table could be a trimmed down version for WASM if you want to bundle it for the browsers. You might want to restrict the number of hashes there to the ones that are most commonly used. Or you might even replace them with some other crates that use the crypto API in browsers. That would be impossible with the code we currently have.

dvc94ch:
Wouldn't it be better to add wasm support with `#[cfg(target_os = "wasm")]` or something like that? The `Code` enum could have some codes disabled when building for wasm. In general the rust community prefers to hide these things like that instead of making it generic and then creating a `multihash-wasm` crate. That makes it harder to support both.

vmx:
Yes, that works for having a "default wasm version" (and it also should be like that). But I'm sure there are cases where someone wants even a more specialized bundle (application specific) and with this approach it's possible. I just wanted to take WASM as an example for the flexibility, not as a definite use case.

Gozala:
> Before this change we had Custom(u64) within the enum. You also couldn't really add a custom hasher to it transparently, you would need to match whether it is a custom code and then return the hasher in your application logic.
> 
> I would argue that custom hashes are rare, hence I'm happy to take the trade-off of being it a bit harder to do, but still possible.

What I'd personally would like to consider if there is the way to improve custom hashes use case without taxing common case. I do not have time budget to look into this in details to offer anything specific, but broadly speaking:

- Can custom hashes be served by a different library that can be used in place of this ? If so that is probably much better route to go about this.
- Is the case common enough to justify added complexity ? If it is that common maybe a better way to solve that is to standardize them ?

vmx:
> What I'd personally would like to consider if there is the way to improve custom hashes use case without taxing common case.

This is exactly what this is about. I don't see how this is taxing the common case.

Gozala:
> > What I'd personally would like to consider if there is the way to improve custom hashes use case without taxing common case.
> 
> This is exactly what this is about. I don't see how this is taxing the common case.

Oh my bad, I must have overlooked the details and was under impression that type parameter was added to Multihash which seemed like added complexity that everyone would share burden of.

I recognize now that this is not the case, and a compromise I was calling for. So please disregard my earlier comment.

vmx:
@dvc94ch Do you still have objections to this PR?

dvc94ch:
I'm not a big fan of having generics everywhere, I opened a PR to remove most of them from rust-ipfs. I think it's a feature that tends to get overused. But in the end your the maintainer of this crate, so do as you think is best.

src/digests.rs
```diff
@@ -8,34 +9,155 @@ use crate::errors::{DecodeError, DecodeOwnedError};
...
+///     ExtendedCode::NormalCode(Code::Sha2_256)
+/// );
+/// assert_eq!(
+///     Code::try_from(ExtendedCode::try_from(0x12).unwrap()).unwrap(),
+///     Code::Sha2_256
+/// );
+/// ```
 #[derive(Clone)]
-pub struct Multihash {
+pub struct MultihashGeneric<T> {
```
dignifiedquire: This should have a trait bound of `TryFrom<u64>` given how much this is being relied upon in the other impls.

src/digests.rs
```diff
@@ -52,15 +174,16 @@ impl Multihash {
...
-        if let Err(err) = MultihashRef::from_slice(&bytes) {
+    pub fn from_bytes(bytes: Vec<u8>) -> Result<MultihashGeneric<T>, DecodeOwnedError> {
+        if let Err(err) = MultihashRefGeneric::<T>::from_slice(&bytes) {
             return Err(DecodeOwnedError {
                 error: err,
                 data: bytes,
             });
         }
-        Ok(Multihash {
+        Ok(Self {
```
dignifiedquire: personal nit: use the actual name

src/digests.rs
```diff
@@ -210,22 +341,25 @@ impl<'a> MultihashRef<'a> {
     /// let mh2 = MultihashRef::from_slice(&mh).unwrap();
     /// assert_eq!(mh2.algorithm(), Code::Sha2_256);
     /// ```
-    pub fn algorithm(&self) -> Code {
-        let (code, _bytes) =
+    pub fn algorithm(&self) -> T
+    where
+        <T as std::convert::TryFrom<u64>>::Error: std::fmt::Debug,
```
dignifiedquire: no need for `std::convert` prefix

src/digests.rs
```diff
@@ -210,22 +341,25 @@ impl<'a> MultihashRef<'a> {
...
+    {
+        let (rawcode, _bytes) =
             varint_decode::u64(&self.bytes).expect("multihash is known to be valid algorithm");
-        Code::from(code)
+        T::try_from(rawcode).unwrap()
```
dignifiedquire: why remove the `expect`?

src/digests.rs
```diff
@@ -312,21 +447,22 @@ pub trait MultihashDigest {
...
+/// use multihash::{wrap, Code, Multihash, Sha2_256};
 ///
 /// let mh = Sha2_256::digest(b"hello world");
 /// let digest = mh.digest();
-/// let wrapped = wrap(Code::Sha2_256, &digest);
+/// let wrapped: Multihash = wrap(Code::Sha2_256, &digest);
 /// assert_eq!(wrapped.digest(), digest);
 /// ```
-pub fn wrap(code: Code, data: &[u8]) -> Multihash {
+pub fn wrap<T: Into<u64> + std::fmt::Debug>(code: T, data: &[u8]) -> MultihashGeneric<T> {
```
dignifiedquire: can we remove this `Debug` bound?

vmx:
@dignifiedquire thanks for the review, my main concern was whether this is generally a good idea or not (it seems to be one).

A few things might still change as I currently try to get rust-cid using this PR and also moving into the same direction. Using this code within rust-cid helps a lot making it cleaner/having better bounds for actual use.

vmx:
@dignifiedquire please review again. I've made a few changes, due to having them actually used in https://github.com/multiformats/rust-cid/pull/50.

Please have a close look at the trait bounds, I don't know much about best practices there.

vmx:
@dignifiedquire I've used `num-traits` before, but then had to drop it again, as `Code` enum had a `Custom(u64)` member. Thanks for bringing it up again, as now it should be possible again.

I also think it makes sense, so this would be a subsequent PR.

rklaehn:
A few quick comments:

- this is a pretty major change. Has this been tested with the major downstream users such as rust-libp2p? If this causes major trouble for them they might just fork it again. We don't want that. Ideally it should just work for them.

- For me personally, I prefer to have the trait bounds only on the methods that require the trait bound, and not on the struct itself. If it does not make sense to create an instance of the thing if you can not satisfy the trait bound, I put the trait bound on the constructor method (e.g. new). I think it is visually distracting to have the trait bound on all trait impls, even those that are completely unrelated to it. E.g. needing `TryFrom` on the Debug instance. But this advice directly contradicts what @dignifiedquire has suggested, so take it as just my opinion.

- I think this significantly complicates the very common usage. Yes, I saw that the generics are hidden behind a type alias. But still, it is not nice to complicate the common case for making a relatively uncommon case easier/better to handle. I do not know how common custom codes are, but I personally have not used them despite being a pretty active user of multiformats, and I would guess that the majority of multiformats users have similar usage patterns.

I don't know enough about the intended use case to suggest a better solution, but it would be great if the complexity could be somehow confined to when you actually need custom code tables, even if this means a bit of copypasta.

vmx:
>     * this is a pretty major change. Has this been tested with the major downstream users such as rust-libp2p? If this causes major trouble for them they might just fork it again. We don't want that. Ideally it should just work for them.

I've tried it with rust-libp2p, rust-cid and rust-ipld. I only needed minor changes (sometimes none).

>     * I think this significantly complicates the very common usage. Yes, I saw that the generics are hidden behind a type alias. But still, it is not nice to complicate the common case for making a relatively uncommon case easier/better to handle. I do not know how common custom codes are, but I personally have not used them despite being a pretty active user of multiformats, and I would guess that the majority of multiformats users have similar usage patterns.

The `MultihashDigest` trait is the only case where I see more complications, as it now needs a generic parameter (as Rust doesn't have aliases for traits yet). For the rest, I think the trait alias works pretty nicely (one thing where it is also different is if you print some debug output, as it contains the original name, e.g. MultihashGeneric).

One use case I already have for custom codecs (to be fair it's not about multihash, but about IPLD codecs, but it is the same idea). Currently rust-cid has its [own codec table](https://github.com/multiformats/rust-cid/blob/fcb8ea3f48eb4eb03c85a373d3e55ba57ce49817/src/codec.rs). If you use IPLD, I'd like to have a table, which only contains the codecs that are also supported for encoding/decoding. So I create a custom table for it. I can pass it into the CID as generic (once https://github.com/multiformats/rust-cid/pull/50 lands) and libraries using IPLD don't need to care about the details. On the other hand, rust-cid doesn't have a dependency on IPLD.

This way we could potentially also have a multihash code table in rust-cid, so rust-cid wouldn't need the full multihash crate as dependency.

rklaehn:
What exactly is the problem with using default type arguments? Those are usually pretty nice for such things. E.g. HashMap has an extra type argument that 95% of people are not even aware of, that allows you to customize it. This seems similar.

vmx:
> What exactly is the problem with using default type arguments?

I can't recall, as the comment says, it has to do with the `PhantomData`. But I can try it again if you like.

vmx:
> I can't recall, as the comment says, it has to do with the `PhantomData`. But I can try it again if you like.

@rklaehn I found again the minimal example that reproduces the issue: 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=dc796d16cbcd6474a953a480b483d5d2

dvc94ch:
The reason why the HashMap works is because of the return type of `new` which is probably `impl<K, V> HashMap { fn new() -> HashMap<K, V, DefaultHasher> }`. Default type parameters are sadly pretty useless as rust tries to infer the type, but doesn't use the default parameter for it's inference.

rklaehn:
I tried the same thing in a slightly different way, and it works. See https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5e0fa7c9eed76f2c79d3f14ae9940b4e

vmx:
@rklaehn I've tried it again with a default parameter, but the problem is that then calls like:

    Multihash::from_bytes(vec![1, 2, 3])

would become:

    Multihash::<Code>::from_bytes(vec![1, 2, 3])

which I think is worse than the current version with a type alias.

What do you think?

vmx:
@rklaehn I've pushed it as a new branch in case you want to play with it: https://github.com/multiformats/rust-multihash/tree/code-generic-default-param

vmx:
I merge this PR. I think I've addressed all open questions. Feel free to open PRs/issues to discuss how we can further improve the API/making things simpler (e.g. getting the generic default parameter work nicely).

PR #63, opened 2020-04-02
feat: implement `hasher()` as `From` to `MultihashDigest<Code>`

vmx:
BREAKING CHANGE: Instead of calling `hasher()` use `from()` instead

Prior to this change, your code looked like this:

    let hasher = Code::Sha3_512.hasher();

Now it is:

    let hasher = Box::<dyn MultihashDigest<Code>>::from(Code::Sha3_512);

I tend to use the (Try)From trait everywhere these days, so please be critical if it doesn't make sense (though in this case I really like it).

### DIFF
--- a/src/arb.rs
+++ b/src/arb.rs
@@ -1,7 +1,7 @@
 use quickcheck::{Arbitrary, Gen};
 use rand::seq::SliceRandom;
 
-use crate::{Code, Code::*, Multihash};
+use crate::{Code, Code::*, Multihash, MultihashDigest};
 
 const HASHES: [Code; 16] = [
     Identity, Sha1, Sha2_256, Sha2_512, Sha3_512, Sha3_384, Sha3_256, Sha3_224, Keccak224,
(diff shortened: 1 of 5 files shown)


vmx:
`MultihashDigest` now has a default generic type and I found out that it infers the type if you use it as `MultihashDigest<_>`. 

But I should explain where the idea for this PR comes from, perhaps there are better ways to solve it.

I don't want Mutlihash code tables to depend on having actual hashing implemented. Use case is rust-cid, where you might want to work with CIDs, but don't ever need to do any hashing.

If you have something like IPLD, you might want to require that your codec table also supports hashing. Currently  the `hasher()` is implemented on the code table enum itself and not on as a trait, so you can't require an implementation to have it implemented. Instead of having a separate trait that implements `hasher()`, I thought I just use a `From` trait instead. Now you can have a trait bound like `Box<dyn MultihashDigest<T>>: From<T>,` where `T` is the code table you want to use.

dignifiedquire:
I wonder if it would make sense to have a nice type alias


```rust
type BoxedMultihashDigest<C> = Box<dyn MultihashDigeset<C>>;

let hasher: BoxedMultihashDigest<_> = Code::Sha3_512.into();
```

There is some precedence for this with things like `BoxedFuture` in the futures library, and I think it makes it more clear that this is meant to be used as a boxed thing.

PR #68, opened 2020-06-12
fix: implement `digest` on the code table as trait

vmx:
Instead of having `digest()` directly on the implementation, implement
it via trait. This way the digest can be used as trait bound.

BREAKING CHANGE: You need to import the `DigestFromCode` trait if you
use `digest()` on a Multihash Code.


vmx:
This PR was triggered by https://github.com/ipfs-rust/rust-ipld/pull/46#discussion_r438155438.

vmx:
Although I plan to have major refactorings, I still think this is a valuable change that makes things cleaner. @dignifiedquire any objections to merging this?

dignifiedquire:
seems reasonable

vmx:
This is now a thing, hence closing this one.

PR #69, opened 2020-06-17
Rewrite multihash.

dvc94ch:
- simplifies the implementation
- no-std support
- proc macro for custom codes
- no allocations

TODO:
- [x] From<MultihashArray<C, S>> for Multihash<C>
  - not trivial to do, would require a new attribute in the proc-macro
- [ ] truncation
  - not trivial to do, if it's possible
- [ ] identity
  - if truncation is implemented it could work using a maximum size and then truncate it to the
    actual length
- [x] rename hasher multihasher trait methods + add reset

@vmx also fixes #68


dvc94ch:
> Improving docs with examples on how to actually use it. It seems that you would usually use the from_bytes() and to_bytes() and the read() and write() are mostly for internal use.

mmh in the ipld cbor codec we want to use read and write to avoid allocations. `from_bytes` and `to_bytes` are utilities to make using it easier, in many cases it's recommended to use `read` and `write` directly, especially when reading or writing from a file.

> Does the Unsigned trait need to be public?

It's needed by the proc-macro to derive `Multihash`.

> In the docs, link Code to the MultihashCode trait, so that it's clear that this value is a trait and not the Code enum.

Not sure what you mean.

dvc94ch:
> Is that a limitation to make things stack allocated, or is an unlimited sized heap allocated identity hash not possible anymore with this code base?

In what scenario does it make sense to have arbitrarily sized identity hashes?

Why truncate a hash? If you want a different hash size you can use the Blake2b, Blake2s or Strobe hashers that support arbitrary digest lengths and derive a custom code table.

vmx:
This is a massive PR, which is a new implementation which makes it difficult to review. You'd need a lot of time. Probably too much time for volunteers to have a look.

It would be great if it could be split up in separate PRs. I know, it's big ask, as it takes a lot of time splitting things up again, but it will save a lot of time for reviewers and hopefully will also attract other reviewers, which will make the code better.

Some ideas how to split it, though those might not work because of how things work:
 - Doing the `GenericArray` things separately, either after or before the API changes.
 - Do the proc-macro in a separate step. Also, is the proc macro really needed because of the limitations of normal macros? The proc macro adds compile time and complexity. Could it be done similarly to what we do atm?

dvc94ch:
The proc macro is needed so that you can use `#[derive(Multihash)]`. The current macro wasn't designed to be used by other crates, as it adds a custom syntax. With the derive proc macro you start by adding `#[derive(Multihash)]` and then the compiler will tell you exactly what to do (add a code attribute, add a module attribute). Error reporting is much better and because we generate a custom `Multihash` enum for the code table, the macro has grown in complexity. Most medium sized projects will have syn/quote in their dependency graph anyway.

vmx:
> > In the docs, link Code to the MultihashCode trait, so that it's clear that this value is a trait and not the Code enum.
> 
> Not sure what you mean.

When you look at the rendered documentation `Code` isn't a link so it is not very clear that it's a trait (that's just a tiny small thing to fix).

> In what scenario does it make sense to have arbitrarily sized identity hashes?

That's a good question. Currently no limitation is specified. Though it might make sense. I wanted to find out if we need to specify a limit in order to make this implementation spec compliant, or if we could (potentially) just have very large identity hashes.

> Why truncate a hash? If you want a different hash size you can use the Blake2b, Blake2s or Strobe hashers that support arbitrary digest lengths and derive a custom code table.

It's a feature of Multihash, that the hash has the size it specifies and not automatically the size of the digest. If you want to know more about that, I can try to find out the reason for that.

> The proc macro is needed so that you can use `#[derive(Multihash)]`.

Oh cool. I didn't see that you can derive `Multihash`. I need to check that out.

vmx:
I forgot to add in regards to separate PRs/commits. Also things like adding a new hash algorithm like Strobe is ideally completely separate. Updating dependencies could also be separate. Those things make it for reviewers easier to see what the change are really about.

vmx:
`MultihashArray` seems to be the central piece. The docs mention:

> Stack allocated multihash storage backend for codes up to 127.

I need support for codes bigger than that. Would that mean that we'd need some additional code to also support codes higher than that (which then might not be stack allocated I guess)?

koivunej:
Definitely agree with vmx on splitting this up.

I wonder how would the downstream impls look after this; could one have a `Vec<Cid>` afterwards or would that have to be vec of some unsizing pointer like box?

Do I also understand correctly that this PR would remove the non-owned story altogether, or is there some slice impl I can't see?

I do see the want to use arrays whenever working with hashes. I wonder if the same end results could had been achieved through making Cid more friendly towards MultihashRef's. Then the hash could reside on the stack, and a non-owned multihash inside a non-owned Cid would be used to produce the final Cid bytes to a writer...? Note I might be lagging on the latest updates on Cid.

dvc94ch:
Ok I'll move it to the ipfs-rust repo, really no point in arguing with you

dvc94ch:
Published as tiny-multihash 0.1.0

vmx:
@dvc94ch I'm sad to see this one closed. It contains lots of good stuff, it really improves the overall API a lot. I had hoped that we can figure out the missing pieces (like supporting truncated hashes).

Stebalien:
@dvc94ch 

> An interesting observation is the @koivunej's shitty code is always good enough to merge. At least @vmx isn't a hypocrite, he actually makes small PR's.

This comment violates the IPFS Code of Conduct.

> We ask that you please respect that people have differences of opinion regarding technical choices, and that every design or implementation choice carries a trade-off and numerous costs. There is seldom a single right answer. A difference of technology preferences is not a license to be rude.

https://github.com/ipfs/community/blob/master/code-of-conduct.md#friendly-harassment-free-space

dvc94ch:
@Stebalien I updated my comment.

Issue #73, opened 2020-07-16
use multibase for encoding/decoding


mriise:
[rust multibase](https://github.com/multiformats/rust-multibase) in tests, could also be a good opportunity to remove some of the brittleness of the tests.

vmx:
I don't quite understand what the core problem is that using multibase would help. You could see the current bytes as Base16 (Multibase Base16 without the prefix).

vmx:
In the context of #74 it makes sense. Using Multibase instead of a custom hex parser. I think for now using the hex crate is simpler than using Multibase.

mriise:
using multibase in tests actually doesn't have much of a point, as its a test of multihash, not multihash _and_ multibase. hex just fine for validating functionality.

PR #74, opened 2020-07-16
Blake3

mriise:
implements trivial support for [blake3](https://github.com/BLAKE3-team/BLAKE3) using the default 32 byte digest.

longer digests could be implemented using XOF with separate codes, or potentially passing in the desired length, neither of which is in this PR though. 

improvements could also be had with [rayon](https://crates.io/crates/rayon) using [join](https://docs.rs/blake3/0.3.5/blake3/join/index.html) but are again not in this PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ edition = "2018"
 [dependencies]
 blake2b_simd = { version = "0.5.9", default-features = false }
 blake2s_simd = { version = "0.5.9", default-features = false }
+blake3 = "0.3.5"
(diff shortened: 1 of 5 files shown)


mriise:
This also removed the hex_to_bytes() helper function in favor of the hex crate, which in retrospect should have instead been the multibase crate. I created a separate issue #73 to discuss that.

mriise:
Eh, code coverage isnt happy with it. would need to add tests for `reset` `result_reset` and IO::Write to get it back up to 71 ish

PR #80, opened 2020-08-26
Optional blake3

### LINKED ISSUE #79: Make CC dependency a (default?) feature
v0.11.3 caused some difficulties for us due to, on some targets, the `blake3` dependency using `cc` to compile intrinsics. 

We maintain a complex cross-platform project that we build for Windows/macOS/Linux/Android/iOS where all C/C++ dependencies are prebuilt and packaged through conan.io packages with strict compilation options. With this in mind, we try to enforce builds where the "cc" crate is never called because of the lack of control over the build process and compilation options. In this specific case, we got an unexpected and sudden build break in our CI (`blake3` is using linker options that are incompatible with ours on Windows x86).

For now, I've pinned 0.11.2 and we can always patch `multihash` ourselves. However, it would be really nice if there was some possibility to just control the hash algorithms available via features. We don't need `blake3` support, and being able to disable it would be more "environmentally friendly". As it stands right now, a C/C++ application built with MSVC for x86 and using /MT won't be able to compile with `multihash` as a dependency. 

Possibly related to #32

mriise:
fixes #79 

313 and 279 in hashes.rs seem a bit lazy to me, but it works fine and compiles down to the same size in the end. If you know of a better way of doing this please lmk, this just required the least amount of changes of actual logic.


Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "multihash"
 description = "Implementation of the multihash format"
 repository = "https://github.com/multiformats/rust-multihash"
 keywords = ["multihash", "ipfs"]
-version = "0.11.3"
+version = "0.11.4"
```
vmx, line 6: Please don't bump the version manually. This is done during the release.

vmx (commented):
Thanks a lot. The duplication is not pretty, but I see it as a temporary fix before a bigger refactoring, so that's fine.

vmx:
@thenextman Could you please try out this PR to see if it fixes your issue?

thenextman:
@vmx I tested it, it works well on our side

vmx (approved):
Thank you @mriise for this PR and @thenextman for verifying that it works as expected.

vmx:
Released as 0.11.4.

PR #82, opened 2020-10-01
Merge tiny-multihash

vmx:
Lots of development has happened at https://github.com/ipfs-rust/tiny-multihash repo (spawned from https://github.com/multiformats/rust-multihash/pull/69). It was rewritten quite a few times, but I think it's now at a state where it can get merged back in again.

It adds `no_std` support as well as allocating digests on the stack only.

For people that have followed the tiny-multihash development, it's easiest for you to just diff between https://github.com/ipfs-rust/tiny-multihash/commit/924b441b0a3c48a52774ae9b246225bfb6974e0b and this PR in order to see that there's no real difference.

For others. I'm sorry that this PR is so massive. You could check the tests to see that not that much on the APIs changed.

@dignifiedquire It would be great to get at a sanity check on this code (as opposed to an in-depth review) to see if you think this is the way to move forward.

One missing feature for feature parity with the current version is BLAKE3 support. That is currently worked on at https://github.com/ipfs-rust/tiny-multihash/pull/36 and can either be finished there and then merged into this repo, or in case this PR gets merged, can directly be moved to here.


vmx:
I totally forgot to thank @dvc94ch for all his work on the code and the original development of stack allocating things and making it `no_std` compatible.

dvc94ch:
To decrease maintenance burden, it would be nice if the cid, multihash code move in to the same repo. That way PRs need to fix any cid changes and releases can be more coordinated. It's happened multiple times that multihash got a new release but cid didn't. If there is a minor version bump a new cid release needs to happen too.

vmx:
As BLAKE3 support landed in in tiny-multihash, I updated this merge and force pushed it.

dvc94ch:
Also you removed Copy from Multihash without mentioning it?

dvc94ch:
Just fyi, I'll continue maintaining tiny-multihash/tiny-cid as I need to update everything else too when there are changes, ensuring that things actually work and not go through months of reviews and release cycles to get fixes in when stuff breaks. Like `Multihash` not being `Copy`.

vmx:
> Also you removed Copy from Multihash without mentioning it?

Oh yes sorry. Removing `Copy` was during development as an automatic derive didn't work (due to GenericArray) and I wanted to make progress. I then I forgot about it. Thanks for bringing it up, I saw you fixed it already on tiny-multihash. I will also put it in here.

> Just fyi, I'll continue maintaining tiny-multihash/tiny-cid as I need to update everything else too when there are changes, ensuring that things actually work and not go through months of reviews and release cycles to get fixes in when stuff breaks. Like `Multihash` not being `Copy`.

Mistakes happen and this is where I see the strength of code review. This is also the reason why I prefer smaller PRs as it's those things that can be caught during code review.

Fixes can also get through quickly if they are small. Like that Copy fix, it won't take months to review.

I really hope that once I've merged the tiny-multihash stuff into this one that you consider making contributions here instead of maintaining tiny-multihash. If things move too slowly then you can always fork again, but I hope that we can work together on this and grow a community.

vmx:
As there was now a new version of tiny-multihash released with an additional fix, I'll redo the merge once more.

dvc94ch:
The reason I assumed it wasn't a mistake is because I saw some wip work on updating rust-cid and libipld, at which point you should have noticed. Sorry for jumping to conclusions.

vmx:
> The reason I assumed it wasn't a mistake is because I saw some wip work on updating rust-cid and libipld, at which point you should have noticed. Sorry for jumping to conclusions.

No worries. Here's the actual story: when I was doing that work I was actually wondering why Copy isn't implemented. I then tried the automatic `Copy` derive, which doesn't work due to `GenericArray`. I then thought "well if GenericArray doesn't implement it, it's probably only possible to clone it" and moved on. It was my lack of Rust knowledge and I still need time to understand why deriving `Copy` manually the way it is, is possible safe and makes sense.

dvc94ch:
Basically all stack allocated types should be Copy according to the docs. Due to rust not supporting const generics we need to help rust a little.

.github/workflows/build.yml
```diff
@@ -63,3 +63,24 @@ jobs:
 
       - name: Upload to codecov.io
         uses: codecov/codecov-action@v1.0.6
+
+  no-default-features:
```
dignifiedquire, line 67: I would suggest adding an actual compile with a `no-std` target, otherwise it is easy to miss things

Cargo.toml
```diff
@@ -1,36 +1,59 @@
...
+members = ["derive", "."]
+
 [package]
 name = "multihash"
 description = "Implementation of the multihash format"
 repository = "https://github.com/multiformats/rust-multihash"
 keywords = ["multihash", "ipfs"]
 version = "0.11.4"
-authors = ["dignifiedquire <dignifiedquire@gmail.com>"]
+authors = ["dignifiedquire <dignifiedquire@gmail.com>", "David Craven <david@craven.ch>"]
```
dignifiedquire, line 10: @vmx I mean at this point you should probably add yourself :D

src/multihash.rs
```diff
@@ -0,0 +1,304 @@
...
+    /// Writes a multihash to a byte stream.
+    #[cfg(feature = "std")]
+    pub fn write<W: std::io::Write>(&self, w: W) -> Result<(), Error> {
+        write_multihash(w, self.code(), self.size(), self.digest())
+    }
+
+    /// Returns the bytes of a multihash.
+    #[cfg(feature = "std")]
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut bytes = vec![];
```
dignifiedquire, line 158: this could be `Vec::with_capacity(self.size())`, to allocate at once

src/multihash_impl.rs
```diff
@@ -0,0 +1,106 @@
+use multihash_derive::Multihash;
+
+/// Default (cryptographically secure) Multihash implementation.
```
dignifiedquire, line 3: maybe remove `(cryptographically secure)`, given the state of Sha1?

dvc94ch: Was sha1 added again? Didn't notice, @vmx and I discussed removing it and it was removed at one point

.github/workflows/build.yml
vmx: I was thinking about using https://github.com/hobofan/cargo-nono. do you mean using that `cargo build --target thumbv6m-none-eabi` "hack"?

src/multihash_impl.rs
vmx: Yes. I'll simply remove Sha1 from the default table.

vmx:
I force-pushed to make the merge commit having a better commit message. I then applied the code review in separate commits.

.github/workflows/build.yml
vmx: I'll do that in a separate PR, dealing with CI is often a lot of noise and I'd really like to get this PR merged :)

dignifiedquire (approved):
nice work!

.github/workflows/build.yml
dignifiedquire: this works pretty well: https://github.com/async-rs/async-std/blob/master/.github/workflows/ci.yml#L83-L99

PR #86, opened 2020-10-13
fix: improve Multihash arbitrary

vmx:
The Multihash arbitrary now works without std. It can also be used without
pulling in all the hashes if it is compiled without the `multihash-impl`
feature.

The feature for enabling the arbitrary was renamed from `test` to `arb`.

BREAKING CHANGE: The `test` feature is now called `arb`.

@rklaehn You originally added the arbitraries, so I hope you've a better understanding
than I have about whether this PR makes sense. Or if we e.g. just do the
`#[cfg(not(feature = "multihash-impl"))]` and remove the other one.

Background is that currently if you would use the `arb` feature, it will pull in all hashers
available as dependency. That's not that good if you just want to be able to use Multihash
arbitraries in your own third party code (like e.g. libp2p).


vmx:
After a second though, I think it's better to keep things simple and not changing the `Arbitrary` for `Multihash` depending on the enabled features. I've opened #87 and close this one.

PR #87, opened 2020-10-19
fix: simplify Multihash arbitrary

vmx:
The Multihash arbitrary now works without std.

BREAKING CHANGE: The feature for enabling the arbitrary was renamed
from `test` to `arb`.


dignifiedquire (commented):
This seems to considerably worsen the quality of the random values, as they are neither known codes anymore, nor actual hashes. Unclear if this is desirable

vmx:
@dignifiedquire there is also https://github.com/multiformats/rust-multihash/pull/86.

Though I'm not sure if it is actually worse. Doesn't it make sense to test with arbitrary values and not just the common ones (I'm not that well versed in quickcheck though).

vmx:
Or would it perhaps make sense to use a custom distribution of the hash code? e.g. having 25% of the random numbers on on of the 1-4 byte ranges? Or having a distribution like 50% 1-byte, 25% 2-byte, 12.5% 3-byte, 12.5% 4-byte?

src/arb.rs
```diff
@@ -1,36 +1,16 @@
...
-        // encoding an actual random piece of data might be better than just choosing
-        // random numbers of the appropriate size, since some hash algos might produce
-        // a limited set of values
-        code.digest(&data)
+        let code = g.gen::<u64>();
+        // Maximum size is 64 byte due to the `U64` generic
+        let size = g.gen_range(0, 64);
+        let mut data = [0; 64];
+        g.fill_bytes(&mut data);
+        Multihash::wrap(code, &data[..size]).unwrap()
```
rklaehn, line 14: What does it mean to have a code in the range from 0..2^64-1?

rklaehn (approved):
Looks reasonable

I wonder if downstream code will be irritated by the unusaully high numbers for code, but then that would be a bug.

An issue with arbitraries is that you want to cover the entire possible value domain, but you also don't want to end up with values that are so obviously absurd that they don't discover "interesting" bugs.

src/arb.rs
rklaehn: You might want to do something where you have plausible code values most frequently, and then just occasionally add a code value from the full possible range.

vmx: So what about changing this to something like I mentioned in my [other comment](https://github.com/multiformats/rust-multihash/pull/87#issuecomment-714404397):

> Or having a distribution like 50% 1-byte, 25% 2-byte, 12.5% 3-byte, 12.5% 4-byte?

I think I'll try that and push another commit.

vmx:
@dignifiedquire @rklaehn does this new version look better now?

dignifiedquire (approved):

src/arb.rs
```diff
@@ -1,36 +1,34 @@
...
-/// This is done by encoding a random piece of data.
 impl Arbitrary for Multihash<U64> {
     fn arbitrary<G: Gen>(g: &mut G) -> Self {
-        let code = *HASHES.choose(g).unwrap();
-        let data: Vec<u8> = Arbitrary::arbitrary(g);
-        // encoding an actual random piece of data might be better than just choosing
-        // random numbers of the appropriate size, since some hash algos might produce
-        // a limited set of values
-        code.digest(&data)
+        // In real workd lower multihash codes are more likely to happen, hence distribute them
```
rklaehn, line 12: s/workd/world/

rklaehn (approved):
Looks reasonable to me. You will only figure out if the distribution is good once you use this for finding bugs, but it certainly is better than a completely random distribution between 0 and 2^64-1, since the latter would basically never hit realistic codec values.

PR #91, opened 2020-11-04
feat: default code table can be changed at compile time

vmx:
It is now possible to change the contents of the default code table
with enabling/disabling various features. If you e.g. only need SHA2
hashes, you can specify that when you use the `multihash` dependency:

    multihash = {
        version = …,
        default-features = false,
        features = ["std", "multihash-impl", "sha2"]
    }

BREAKING CHANGE: Only secure hashers are enabled by default

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,17 +14,18 @@ documentation = "https://docs.rs/multihash/"
 edition = "2018"
 
 [features]
-default = ["std", "all", "derive", "multihash-impl"]
+default = ["std", "derive", "multihash-impl", "secure-hashes"]
 std = ["unsigned-varint/std", "multihash-derive/std"]
-multihash-impl = ["derive", "all"]
+multihash-impl = ["derive"]
 derive = ["multihash-derive"]
 arb = ["quickcheck", "rand"]
-all = ["blake2b", "blake2s", "blake3", "sha1", "sha2", "sha3", "strobe"]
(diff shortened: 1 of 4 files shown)


dignifiedquire (approved):

Cargo.toml
```diff
@@ -14,17 +14,18 @@ documentation = "https://docs.rs/multihash/"
...
 [features]
-default = ["std", "all", "derive", "multihash-impl"]
+default = ["std", "derive", "multihash-impl", "secure-hashes"]
 std = ["unsigned-varint/std", "multihash-derive/std"]
-multihash-impl = ["derive", "all"]
+multihash-impl = ["derive"]
 derive = ["multihash-derive"]
 arb = ["quickcheck", "rand"]
-all = ["blake2b", "blake2s", "blake3", "sha1", "sha2", "sha3", "strobe"]
+secure-hashes = ["blake2b", "blake2s", "blake3", "sha2", "sha3"]
```
mriise, line 22: what is meant by secure? the blake family aren't cryptographically secure as far as I know.

dignifiedquire: they are secure in the sense of cryptographic hash functions: https://en.wikipedia.org/wiki/Cryptographic_hash_function which are different from hash functions suitable for password hashing. 
Password Hash functions usually need to be KDF (key derivation function) and be expensive to evaluate. Some more details eg here: https://crypto.stackexchange.com/a/70722

PR #92, opened 2020-11-09
feat: create Multihash that matches the code table

vmx:
Instead of only export the `Multihash` that is generic over the allocated
size, also expose `Multihash` that is created by the `Multihash` derive.

If the `multihash-impl` feature is enabled, also a `Multihash` with the
corresponding size gets exported.

BREAKING CHANGE: The `Multihash` which is generic over the size is now
called `MultihashGeneric`.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -1,4 +1,5 @@
-//! This proc macro derives a custom Multihash code table from a list of hashers.
+//! This proc macro derives a custom Multihash code table from a list of hashers. It also
+//! generates a public type called `Multihash` which corresponds to the specified `alloc_size`.
 //!
 //! The digests are stack allocated with a fixed size. That size needs to be big enough to hold any
 //! of the specified hash digests. This cannot be determined reliably on compile-time, hence it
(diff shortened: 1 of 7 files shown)


vmx:
@dvc94ch it would be great to get your feedback on this one. This was triggered by the review of my [libp2p multihash upgrade PR](https://github.com/libp2p/rust-libp2p/pull/1792). I think it's a clean way of making the library usable by cases like libp2p, but still having full flexibility (the cases I had in mind when I was working with you on that). I also think it still matches well with `derive(Multihash)` that a `Multihash` with a specific size is derived.

dignifiedquire (commented):
I like it

dvc94ch:
Is this just about removing the generic via type aliasing? That seems ok to me.

vmx:
> Is this just about removing the generic via type aliasing? That seems ok to me.

Yes it does with the derive. If you need the generic type you can e.g. do `use multihash::MultihashGeneric as Multihash` and then use the `Multihash<U64>` as you're used to.

PR #93, opened 2020-11-16
feat: derive Hash for Mutlihash<S>

vmx:


src/multihash.rs
```diff
@@ -74,7 +74,7 @@ pub trait MultihashDigest:
 #[cfg_attr(feature = "serde-codec", derive(serde::Deserialize))]
 #[cfg_attr(feature = "serde-codec", derive(serde::Serialize))]
 #[cfg_attr(feature = "serde-codec", serde(bound = "S: Size"))]
-#[derive(Clone, Debug, Default, Eq, Ord, PartialEq, PartialOrd)]
+#[derive(Clone, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
austinabell, line 77: Technically this is hashing more data than necessary since it would hash the size and the full allocated array (even if the hash uses less), should only need to hash the code and slice of digest based on the size, if you care to micro optimize

vmx: Good point. While I'm at it, I'll just do that.

dignifiedquire (approved):

PR #97, opened 2021-01-13
feat: implement std::io::Write trait for the hashers

### LINKED ISSUE #95: WriteHasher is unusable
multihash 0.13 has introduced `WriteHasher` as a newtype wrapper to provide a generic implementation of `Write` for all hashers.
However, the crate provides no way to construct an instance of `WriteHasher` or obtain a digest from the hasher state.

vmx:
Instead of having a (non-working) wrapper struct for implementing
std::io::Write for a hasher, implement it directly on the hasher
instead.

Closes #95.

@dvc94ch you implemented the original `WriteHasher` wrapper. Is there a reason to do it that way or was it just a different code base and this PR is a better implementation?

### DIFF
--- a/derive/src/multihash.rs
+++ b/derive/src/multihash.rs
@@ -266,7 +266,7 @@ fn error_alloc_size(hashes: &[Hash], expected_alloc_size_type: &syn::Type) {
 
     let maybe_error: Result<(), ParseError> = hashes
         .iter()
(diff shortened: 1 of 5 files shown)


dvc94ch (approved):

vmx:
Thanks for the review!

PR #104, opened 2021-01-16
chore: improve github actions

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,6 @@
 name: build
 
-on:
-  push:
-    branches: [master]
-  pull_request:
+on: [push, pull_request]
 
 jobs:
   build:
@@ -22,9 +19,7 @@ jobs:
       - name: Cache Dependencies & Build Outputs
         uses: actions/cache@v2
         with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
+          path: ~/.cargo
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Install Rust Toolchain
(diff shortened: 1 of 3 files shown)


koushiro:
@vmx PTAL

.github/workflows/build.yml
```diff
@@ -1,85 +1,93 @@
...
+          target: wasm32-unknown-unknown
+      - name: Build
+        uses: actions-rs/cargo@v1
         with:
-          args: '--all-features -- --test-threads 1'
-
-      - name: Upload to codecov.io
-        uses: codecov/codecov-action@v1.0.6
+          command: build
+          args: --no-default-features --target wasm32-unknown-unknown
```
vmx, line 71: This one fails for me locally, due to `core` not being there.

Based on https://blog.dbrgn.ch/2019/12/24/testing-for-no-std-compatibility/ `thumbv6m-none-eabi` should work for ensuring `no_std` compatibility.

vmx (commented):
This PR removes running clippy with `--no-default-features`, please add that again.

.github/workflows/build.yml
vmx: The code from https://github.com/multiformats/rust-multibase/blob/c87cf2bafbd1e808ed75bb9e7ebf6d80b023fd33/.github/workflows/build.yml#L43-L56 can be used.

vmx:
I've merged https://github.com/multiformats/rust-multihash/pull/97, so you shouldn't need the change to make clippy happy anymore.

vmx (commented):
Thanks a lot!

PR #120, opened 2021-04-30
*: Remove parity-scale-codec

mxinden:
Given that there are no known users of the scale encoding feature, this
commit removes it, making overall maintenance easier.

For more details see:
https://github.com/multiformats/rust-multihash/pull/118

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,6 @@ multihash-impl = ["derive"]
 derive = ["multihash-derive"]
 arb = ["quickcheck", "rand"]
 secure-hashes = ["blake2b", "blake2s", "blake3", "sha2", "sha3"]
-scale-codec = ["parity-scale-codec"]
 serde-codec = ["serde", "generic-array/serde"]
 
 blake2b = ["blake2b_simd"]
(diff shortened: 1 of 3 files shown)


mxinden:
Note that this breaks `rust-cid`, see https://github.com/multiformats/rust-cid/pull/72.

rklaehn:
So does parity use parity-scale-codec in rust-cid? I am always in favour of removing things, but if parity uses parity-scale-codec via rust-cid, I guess we should not do it...

dvc94ch:
Someone might complain. But parity doesn't use it directly, if anything one of their customers (people who buy dot).

mxinden:
@tomaka would you mind dropping this in some Parity internal channel to give it more visibility?

In case no one objects I would suggest merging in a week. In the unlikely case where someone still needs it, but only finds out after a release, we can still revert it.

dvc94ch:
I think a guy from social network or whatever it's called (original name for a social network) sometimes posts on the rust-ipfs riot mentioned he was using it.

vmx:
Given that there are peiple using the parity-scale-codec in `rust-cid` and care about it (see https://github.com/multiformats/rust-multihash/pull/128) I think we should close this issue.

PR #127, opened 2021-05-26
Bump proc-macro-crate to v1.0.0

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>

Close #112 

Also close useless PRs (see dependabot/dependabot-core#3704):

Close #122 
Close #123 
Close #124 
Close #125 
Close #126

### DIFF
--- a/derive/src/multihash.rs
+++ b/derive/src/multihash.rs
@@ -7,7 +7,6 @@ use quote::quote;
 use quote::ToTokens;
 use syn::parse::{Parse, ParseStream};
 use syn::spanned::Spanned;
-use syn::Error;
 use synstructure::{Structure, VariantInfo};
 
 mod kw {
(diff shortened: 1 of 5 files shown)


koushiro:
@vmx PTAL

derive/src/multihash.rs
```diff
@@ -314,7 +315,13 @@ fn error_alloc_size(hashes: &[Hash], expected_alloc_size_type: &syn::Type) {
 }
 
 pub fn multihash(s: Structure) -> TokenStream {
-    let mh_crate = utils::use_crate("multihash");
+    let mh_crate = match utils::use_crate("multihash") {
+        Ok(ident) => ident,
+        Err(e) => {
+            let err = Error::new(Span::call_site(), e).to_compile_error();
```
vmx, line 321: As we have different errors in this file, I find it clearer if you would use `syn::Error` instead of just `Error` here. Can you please change that?

derive/src/multihash.rs
```diff
@@ -1,12 +1,13 @@
 use std::collections::HashSet;
 
 use crate::utils;
-use proc_macro2::TokenStream;
+use proc_macro2::{Span, TokenStream};
```
vmx, line 4: As you're now importing `Span`, can you please change the other occurrence in the code from `proc_macro2::Span` to just `Span`?

vmx (commented):
For future changes, I prefer if changes are self-contained. E.g. the change to the CI file or the `.gitignore` is not related to this change, hence should be a separate PR (or at least a separate commit).

derive/src/multihash.rs
koushiro: Done

PR #138, opened 2021-09-14
WIP: allow read/write without std

Stebalien:
Unfortunately, there is no core::io. So this implementation works by duplicating several functions and implementing them for slices.

### DIFF
--- a/src/multihash.rs
+++ b/src/multihash.rs
@@ -127,6 +127,16 @@ impl<S: Size> Multihash<S> {
         Ok(Self { code, size, digest })
     }
 
+    /// Reads a multihash from a slice of bytes.
+    #[cfg(not(feature = "std"))]
+    pub fn read(r: &mut &[u8]) -> Result<Self, Error>
+    where
+        Self: Sized,
+    {
+        let (code, size, digest) = read_multihash(r)?;
+        Ok(Self { code, size, digest })
+    }
+
     /// Parses a multihash from a bytes.
     ///
     /// You need to make sure the passed in bytes have the correct length. The digest length
@@ -150,6 +160,12 @@ impl<S: Size> Multihash<S> {
     /// Writes a multihash to a byte stream.
     #[cfg(feature = "std")]
     pub fn write<W: std::io::Write>(&self, w: W) -> Result<(), Error> {
+        write_multihash(w, self.code(), self.size(), self.digest())
+    }
+
+    /// Writes a multihash to a byte slice, updating the byte slice.
+    #[cfg(not(feature = "std"))]
+    pub fn write<'a>(&self, w: &'a mut &'a mut [u8]) -> Result<(), Error> {
         write_multihash(w, self.code(), self.size(), self.digest())
     }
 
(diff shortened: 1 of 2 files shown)


Ericson2314:
> Unfortunately, there is no core::io

Amen to that...

src/multihash.rs
```diff
@@ -259,6 +275,37 @@ where
...
+    size: u8,
+    digest: &[u8],
+) -> Result<(), Error> {
+    use unsigned_varint::encode as varint_encode;
+
+    let mut code_buf = varint_encode::u64_buffer();
+    let code = varint_encode::u64(code, &mut code_buf);
+
+    let mut size_buf = varint_encode::u8_buffer();
+    let size = varint_encode::u8(size, &mut size_buf);
```
vmx, line 292: This part could be re-used for both, the std and non-std implementation. I'm not sure if it makes the code easier to read, but it might be worth a try.

src/multihash.rs
```diff
@@ -259,6 +275,37 @@ where
...
+
+    let result_size = size.len() + code.len() + digest.len();
+    if result_size > w.len() {
+        return Err(Error::InvalidSize(result_size as u64)); // not the right error.
```
vmx, line 296: Perhaps it would make sense to introduce a non-std `Io` error, which could be used here? I'm not sure myself, perhaps it even warrants a completely new error.

vmx:
I think a rebase might get the CI back to green.

src/multihash.rs
```diff
@@ -259,6 +275,37 @@ where
...
+    }
+
+    w.copy_from_slice(code);
```
mxinden, line 299: Both `src` and `dst` slice need to be of same length, right? This would not be the case here and in consecutive calls.

https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.copy_from_slice

src/multihash.rs
```diff
@@ -285,6 +332,37 @@ where
...
 }
 
+/// Reads a multihash from a byte slice that contains a full multihash (code, size and the digest)
+///
+/// Returns the code, size and the digest. The size is the actual size and not the
+/// maximum/allocated size of the digest.
+///
+/// Currently the maximum size for a digest is 255 bytes.
+#[cfg(not(feature = "std"))]
+pub fn read_multihash<S>(r: &mut &[u8]) -> Result<(u64, u8, GenericArray<u8, S>), Error>
```
mxinden, line 342: Why is a _double pointer_ needed here?

src/multihash.rs
```diff
@@ -285,6 +332,37 @@ where
...
+    let buf = *r;
+
+    let (code, buf) = decode::u64(buf)?;
+    let (size, buf) = decode::u64(buf)?;
+
+    if size > S::to_u64() || size > u8::MAX as u64 || size > buf.len() as u64 {
+        return Err(Error::InvalidSize(size));
+    }
+
+    *r = buf;
```
mxinden, line 357: I guess this reassignment is again related to the _double pointer_ argument? I find this to be an unintuitive side effect. Can this be made more explicit?

Stebalien:
(closing because using slices this way isn't going to be user friendly)

PR #148, opened 2021-10-29
implement const generics

### LINKED ISSUE #114: Use const generics
They are finally there: https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html

Should help a lot with the very verbose types we currently use...

Stebalien:
Rebased and squashed from #116.

Closes #114.

### DIFF
--- a/src/arb.rs
+++ b/src/arb.rs
@@ -7,7 +7,7 @@ use rand::{
 use crate::MultihashGeneric;
 
 /// Generates a random valid multihash.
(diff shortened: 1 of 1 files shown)


koushiro:
> @koushiro You are the only Scale Codec user I know of. Would you mind having a look if things still work as expected. Would you be up for adding tests for the Scale Codec in a separate PR? I always fear we break that part accidentally.

Sorry for the late reply, the scale-codec part LGTM

PR #154, opened 2021-11-04
Support multihash resizing

Stebalien:
TODO: should _truncate_ mutate or return a new object? Currently, I'm returning a new object.

### DIFF
--- a/src/multihash.rs
+++ b/src/multihash.rs
@@ -190,8 +190,8 @@ impl<const S: usize> Multihash<S> {
         mh
     }
 
-    /// Resizes the backing multihash buffer. If the multihash doesn't fit, this function fails and
-    /// does not truncate. If you need to truncate, call that first.
+    /// Resizes the backing multihash buffer. This function fails if the hash digest is larger than
+    /// the target size.
     ///
     /// ```
     /// use multihash::{Code, MultihashDigest, MultihashGeneric};
--- a/src/multihash.rs
+++ b/src/multihash.rs
@@ -173,6 +173,45 @@ impl<const S: usize> Multihash<S> {
             .expect("writing to a vec should never fail");
         bytes
     }
+
+    /// Truncates the multihash to the given size. It's up to the caller to ensure that the new size
+    /// is safe to use.
+    ///
+    /// If the new size is larger than the current size, this method does nothing.
+    ///
+    /// ```
+    /// use multihash::{Code, MultihashDigest};
+    ///
+    /// let hash = Code::Sha3_256.digest(b"Hello world!").truncate(20);
+    /// ```
+    pub fn truncate(&self, size: u8) -> Self {
+        let mut mh = *self;
+        mh.size = mh.size.min(size);
+        mh
+    }
+
+    /// Resizes the backing multihash buffer. If the multihash doesn't fit, this function fails and
+    /// does not truncate. If you need to truncate, call that first.
+    ///
+    /// ```
+    /// use multihash::{Code, MultihashDigest, MultihashGeneric};
+    ///
+    /// let hash = Code::Sha3_256.digest(b"Hello world!");
+    /// let large_hash: MultihashGeneric<32> = hash.resize().unwrap();
+    /// ```
+    pub fn resize<const R: usize>(&self) -> Result<Multihash<R>, Error> {
+        let size = self.size as usize;
+        if size > R {
+            return Err(Error::InvalidSize(self.size as u64));
+        }
+        let mut mh = Multihash {
+            code: self.code,
+            size: self.size,
+            digest: [0; R],
+        };
+        mh.digest[..size].copy_from_slice(&self.digest[..size]);
+        Ok(mh)
+    }
 }
 
 // Don't hash the whole allocated space, but just the actual digest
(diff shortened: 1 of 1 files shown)


dignifiedquire (commented):
I can haz tests?

Stebalien:
Yes....

Stebalien:
Well, resize can't (it changes the type). But yeah, I agree about truncate.

Stebalien:
Hm. Actually, no. This just doesn't behave like a vector.  One will usually want to call something like `let my_digest = my_code::digest("data").truncate(20)`.

src/multihash.rs
```diff
@@ -173,6 +173,45 @@ impl<const S: usize> Multihash<S> {
...
+    ///
+    /// let hash = Code::Sha3_256.digest(b"Hello world!").truncate(20);
+    /// ```
+    pub fn truncate(&self, size: u8) -> Self {
+        let mut mh = *self;
+        mh.size = mh.size.min(size);
+        mh
+    }
+
+    /// Resizes the backing multihash buffer. If the multihash doesn't fit, this function fails and
```
vmx, line 193: I didn't get what "doesn't fit" means when I read the comment. It fails, when the multihash is larger then the new size. Perhaps it can be rephrased?

vmx (commented):
Could you please also add a test case for the error cases?

src/multihash.rs
```diff
@@ -308,6 +347,38 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_resize_up() {
+        let hash = Code::Sha2_256.digest(b"hello world");
+        let _: Multihash<100> = hash.resize().unwrap();
+    }
+
+    #[test]
+    fn test_resize_truncate() {
```
Stebalien, line 377: This is the test for the only error case I'm aware of. Am I missing something?

src/multihash.rs
Stebalien: LGTY?

vmx (approved):
I approve, but I'd like (if possible) also @dignifiedquire to approve it before it is merged.

src/multihash.rs
```diff
@@ -173,6 +173,45 @@ impl<const S: usize> Multihash<S> {
...
+    /// Resizes the backing multihash buffer. This function fails if the hash digest is larger than
+    /// the target size.
+    ///
+    /// ```
+    /// use multihash::{Code, MultihashDigest, MultihashGeneric};
+    ///
+    /// let hash = Code::Sha3_256.digest(b"Hello world!");
+    /// let large_hash: MultihashGeneric<32> = hash.resize().unwrap();
+    /// ```
+    pub fn resize<const R: usize>(&self) -> Result<Multihash<R>, Error> {
```
mriise, line 202: This should ideally be a compiler error instead of a runtime error as we should know both the digest size and the resize size(?) at compile time. the heapless crate has a [neat hack](https://github.com/japaric/heapless/blob/master/src/sealed.rs) to make the compiler notice this in stable rust

mriise: A playground link demonstrating https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=47e71bb4e1fbe3dbd24ac182918c1d93

src/multihash.rs
```diff
@@ -173,6 +173,45 @@ impl<const S: usize> Multihash<S> {
...
+    /// is safe to use.
+    ///
+    /// If the new size is larger than the current size, this method does nothing.
+    ///
+    /// ```
+    /// use multihash::{Code, MultihashDigest};
+    ///
+    /// let hash = Code::Sha3_256.digest(b"Hello world!").truncate(20);
+    /// ```
+    pub fn truncate(&self, size: u8) -> Self {
```
mriise, line 187: since its up to the caller to ensure the size is safe to use, this should be marked `unsafe`. while not quite UB as somewhere it would probably error, Sha2_512 truncated into 32 bytes is something that isnt exactly great. A side note for this, we arent doing any checks to ensure the size of what we read matches with the expected size of the hash, which I will open probably an issue for.

mriise: actually the issue I stated above is currently avoided as we dont have any way of converting a multihash into the digest type. Though reading and creation of malformed multihashes from external sources is allowed.

src/multihash.rs
Stebalien: We know the buffer size, but not the _digest_ size. It should be possible to resize from a `Multihash<64>` to a `Multihash<32>` if the digest fits.

src/multihash.rs
Stebalien: "unsafe" is for rust-level unsafety. The new digest may not be secure, but not all of our hash functions are "secure" anyways. That kind of security needs to be enforced through a list of allowed hash functions/sizes.

mriise: I was not talking about the security of the hash functions. What I am talking about is the allowance of Multihash structs to be formed from non-spec conforming data, every function I saw that converts bytes into a multihash performs no checks outside of if it fits within buffer and struct bounds. 
The unsafe keyword is a good way to inform the user that using this can easily cause malformed data and/or UB if not properly handled, although this wont cause any memory errors or UB since its all safe rust, easily produces incorrect data if used incorrectly and because of that I think it should be labeled `unsafe`.
The `size` field is used to read and write multihashes stuff like the [example here](https://github.com/multiformats/rust-multihash/issues/157) is considered a valid multihash going in and out which is not good.

IMO at the very least a better wording of this function would be `truncate_unchecked` to at least express the need for the user to check what they are doing or at least read the documentation.

mriise: really this should all be checked when decoding to prevent malformed data which is why I created an issue, but since its not checked when encoding either, this function can/will allow for malformed Multihashes to be produced by this lib.

And again, there is no risk of UB or segfaults here, just the risk of outputting multihashes that dont make sense.

src/multihash.rs
mriise: we should know the digest size! even with blake3 we limited it to 32 bytes to ensure we know how big it is.

dignifiedquire, line 202: I wonder if grow would be a more fitting name, given that it can only be larger than the original

dignifiedquire (approved):

src/multihash.rs
vmx: I like the idea of having it called `truncate_unchecked`. I prefer if libraries try hard to produce valid data.

Stebalien: This isn't unchecked and it always produces _valid_ multihashes. It just might not produce _secure_ hash digests.

> The unsafe keyword is a good way to inform the user that using this can easily cause malformed data and/or UB if not properly handled, although this wont cause any memory errors or UB since its all safe rust, easily produces incorrect data if used incorrectly and because of that I think it should be labeled unsafe.

Please don't use unsafe for this. That's not what it means and it makes code difficult to read/audit.

Stebalien: I have updated the documentation to clarify the safety (safe -> secure).

Stebalien: > What I am talking about is the allowance of Multihash structs to be formed from non-spec conforming data

To be clear, the multihash spec allows multihashes to be arbitrarily truncated. I.e., a length-20 SHA256 multihash is a SHA256 multihash truncated to 20 bytes.

vmx (approved):

Issue #141, opened 2021-09-28
No elegent way to stream hash given a hash code


CBenoit:
Hi,

In previous `multihash` version, we used to be able to compute the digest in a streamed manner using [`MultihashDigest::input`](https://docs.rs/multihash/0.11.4/multihash/trait.MultihashDigest.html#tymethod.input) and it was possible to get [a boxed MultihashDigest given a multihash](https://docs.rs/multihash/0.11.4/multihash/type.BoxedMultihashDigest.html).
I currently see no way of doing the same, which is an issue in some use cases.

For example, I need to validate a digest computed from a file. Since the file can be big, I want to use the new `StatefulHasher` trait. However, I found no way to get a trait object.

Here’s my code:

```rust
pub fn validate_file_checksum(expected_digest: &str, file_path: &Path) -> std::io::Result<bool> {
    let (_, hash_data) = multibase::decode(expected_digest).map_err(|e| Error::new(ErrorKind::InvalidInput, e))?;
    let expected_digest = Multihash::from_bytes(&hash_data).map_err(|e| Error::new(ErrorKind::InvalidInput, e))?;
    let hash_code = multihash::Code::try_from(expected_digest.code()).map_err(|e| Error::new(ErrorKind::InvalidInput, e))?;

    // FIXME: multihash new API is breaking this code for streaming hashing (checked for version 0.14)
    //
    //const BUF_SIZE: usize = 1024 * 128;
    //let file = File::open(file_path)?;
    //let mut reader = BufReader::with_capacity(BUF_SIZE, file);
    //
    //let hasher = todo!("get an appropriate trait object hasher given the hash code");
    //
    //loop {
    //    let length = {
    //        let buffer = reader.fill_buf()?;
    //        hasher.update(buffer);
    //        buffer.len()
    //    };
    //    if length == 0 {
    //        break;
    //    }
    //    reader.consume(length);
    //}
    //
    //let digest_found = hasher.finalize();
    //
    // So instead, we read the whole file in memory:

    let file_content = std::fs::read_to_string(file_path)?;
    let digest_found = hash_code.digest(file_content.as_bytes());

    Ok(expected_digest == digest_found)
}
```

If I overlooked something, please let me know!

Thank you

mriise:
It is a bit confusing as both Hasher and StatefulHasher implement Default, but you are explicit about what you want Rust will give it to you.
```rust
let hasher: StatefulHasher = Identity256::default();
```
hopefully this works for you :)

[...]

PR #156, opened 2021-11-04
remove per-hash digest type

Stebalien:
Remove the per-hasher digest type. Instead, store hash digests inside the hashers and "borrow" it. In all cases, we're going to copy it into a `Multihash<S>` anyways.

This:

1. Removes bunch of code.
2. Means that hashers don't need to be generic over the size (unless they actually support multiple sizes). This fixes the UX issue introduced in the const generics PR.
3. Avoids some copying.

BREAKING CHANGE

1. `Hasher.digest` no longer exists. Users should use `Code::SomeCode.digest` where possible.
2. The hasher digests no longer exist.

TODO: Do we want this? From what I can tell, most users won't really be impacted by it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,9 +57,9 @@ use multihash::MultihashCode;
 #[derive(Clone, Copy, Debug, Eq, Multihash, PartialEq)]
 #[mh(alloc_size = 64)]
 pub enum Code {
-    #[mh(code = 0x01, hasher = multihash::Sha2_256, digest = multihash::Sha2Digest<32>)]
+    #[mh(code = 0x01, hasher = multihash::Sha2_256)]
     Foo,
-    #[mh(code = 0x02, hasher = multihash::Sha2_512, digest = multihash::Sha2Digest<64>)]
+    #[mh(code = 0x02, hasher = multihash::Sha2_512)]
     Bar,
 }
 
--- a/benches/multihash.rs
+++ b/benches/multihash.rs
@@ -4,7 +4,7 @@ use rand::Rng;
 use multihash::{
     Blake2b256, Blake2b512, Blake2s128, Blake2s256, Blake3_256, Hasher, Keccak224, Keccak256,
     Keccak384, Keccak512, Sha1, Sha2_256, Sha2_512, Sha3_224, Sha3_256, Sha3_384, Sha3_512,
-    StatefulHasher, Strobe256, Strobe512,
+    Strobe256, Strobe512,
 };
 
 macro_rules! group_digest {
@@ -13,7 +13,9 @@ macro_rules! group_digest {
         $(
             group.bench_function($id, |b| {
                 b.iter(|| {
-                    let _ = black_box($hash::digest($input));
+                    let mut hasher = $hash::default();
+                    hasher.update(black_box($input));
+                    let _ = black_box(hasher.finalize());
                 })
             });
         )*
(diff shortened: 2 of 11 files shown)


vmx (commented):
I like how much simpler the code becomes.

Though if I understand this change correctly, you no longer have the guarantee that you create the expected Multihash from a hasher. To explain what I mean, take this examples:

```rust
let mut hasher = Sha3_256::default();
hasher.update(b"Hello world!");
let hash = Code::Sha2_256.wrap(&hasher.finalize());
```

This won't be a compile-time error, not even a run-time error. You accidentally end up with the wrong Multihash (SHA2 vs. SHA3). With the current code, you'll always get the correct SHA3 Multihash.

Stebalien:
Yes, but I expect users will _usually_ use `Code::SomeHash.digest`. Basically, I'm separating the "multihash" part of this library and the "hasher" part.

I'm also working on a patch to fix #141 which should allow streaming as well. I.e., you'd call `Code::SomeHash.digest_reader(...)`.

Stebalien:
See https://github.com/multiformats/rust-multihash/pull/159 for hashing from a stream.

dignifiedquire:
I like simplifications, but it removes a level of typesafety that I find very useful. In addition the Digest type has been used in all the rustcrypto hashers and that works very nicely (and is what a large part of the ecosystem expects).

Stebalien:
From my perspective, it:

1. Mixes concerns. The "hasher" layer and "multihash" are independent.
2. Prevents hashers from being object safe and makes them harder to work with.

> I like simplifications, but it removes a level of typesafety that I find very useful.

Can you give an example? I can't think of a case where I'd rather manually use a hasher then wrap it in a multihash instead of just calling `Code::HashCode.digest`.

dignifiedquire:
when I have an api and I want to enforce a specific hash digest as input, how can I do this with the new construction?

Stebalien:
> when I have an api and I want to enforce a specific hash digest as input, how can I do this with the new construction?

Is that really in scope? Multihash is all about abstracting over hash functions.

I guess my question is, do you have a concrete case where you needed this? In practice, I'd expect to want to specify something like "I accept a _multihash_ (not just a raw digest) with some set of constraints" (which isn't something we're going to be able to support without dependent typing).

dignifiedquire:
To be fair, I don’t use multihash atm, so I don’t have a particular use case. Really I would allmost never use the hashing part of multihash and hardcode hash digests/be generic over them until I have to interact with the outside world. In my experience being entirely generic over hashfunctions is really not great.

vmx:
> Really I would allmost never use the hashing part of multihash

I'm not sure I fully understand. Do you mean that you would use some hashing within your system and then at the end wrap it into a Multihash? So you don't need the `digest()` or `digest_reader()` calls?

If that's the case, I think then this PR makes sense. I agree with @Stebalien that with the introduction of streaming hashers (#159), I think you would hardly ever use any hashers directly. So making the code simpler makes sense to me. The central entry point for this library is the generated `Code` enum.

mriise:
Ive been doing some work recently to try and remove the proc macro altogether and use an enum with a trait impl to do the work of the macro and it will still require having extra traits to define the size of the array without using unstable rust.

note: this is more of an aside in case it was relevant than saying anything about this PR

vmx:
I've merged it into one commit and rebased it on top of master. @Stebalien you can't review this PR as it is your's, a thumbs up would be good anyway.

@dignifiedquire Do you still have objections that would prevent merging this PR? It seems like @mriise @Stebalien and myself are in favour of making this change.

vmx (approved):

Stebalien (commented):
no objections to merging

mriise (approved):

PR #177, opened 2022-01-25
Several improvements

vmx:
I'm usually a fan of having a clean Git history. Though in this case I think we want all the changes from the `next` branch to land on master. Also all changes were reviewed before they were merged into the `next` branch. So I'm in favour of merging the `next` branch into `master` now.

There are still two PRs open that I think make sense to land before we cut a release (they probably need to be polished a bit more). But this can be done after this one is merged.

It would be great if everyone involved (@dignifiedquire, @mriise, @samuelburnham, @Stebalien) in the changes could give a thumbs up for this PR to land on `master`. (as it really is a lot of (breaking) changes, most notably @mriise excellent long running effort on making const generics work).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,15 +28,9 @@ checksum = "a4c527152e37cf757a3f78aae5a06fbeefdb07ccc535c980a3208ee3060dd544"
 
 [[package]]
 name = "arrayvec"
-version = "0.5.2"
+version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
-
-[[package]]
-name = "arrayvec"
-version = "0.7.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "be4dc07131ffa69b8072d35f5007352af944213cde02545e2103680baed38fcd"
+checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
 name = "atty"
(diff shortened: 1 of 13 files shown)


mriise:
Thanks for the ping @vmx! I would like to hold of just a bit longer. Rust version 1.59 will be introducing const generic defaults which will allow for better ergonomics when dealing with different sized arrays by expanding (or potentially shrinking) the backing array. Users would still need to ensure the Digest type is sized properly but we can add lost of tools to make it easy.

here is a playground link:
https://play.rust-lang.org/?version=beta&mode=debug&edition=2021&gist=138ba86ee23c866228af859d86b7d4ed

There is a downside for this approach however is the const assert hack is very much a hack and spits out potentially confusing error information if types are not what it expected. However I feel that with good documentation, it would be worth for the less confusing interface of types in general.

vmx:
@mriise Do I understand correctly, that your proposed changes would change the current API around resizing Multihashes. As this isn't a feature in any released version yet, I don't expect many people using it already. As @Stebalien introduced that change, I guess that he needs that change. Though I don't expect, that it would be a problem to have a breaking change once Rust 1.59 is out.

@mriise would it work for you if we merge the current version (and have a release), but you're free to break the resizing API as soon as Rust 1. (@Stebalien would need to agree on this too).

To explain my rush: rust-multihash is a dependency deep down, and I'd like to update other libraries like rust-cid and libipld to also use this const generics version.

mriise:
@vmx sorry for the confusion my suggestion had 2 parts, the primary being to change the digest so that it uses a default size and the other to help with resizing the backing array without pulling the digest out of our implementation and putting it back in to fit the generic size. adding an `into_generic()` along with exposing the methods that it uses to resize.

using a default size for the backing array would be doing exactly what we are doing currently to avoid specifying what size each Digest's backing uses (addressing the whole concern about loss in usability). I will make a PR to `next` to show what I mean.

vmx:
> using a default size for the backing array would be doing exactly what we are doing currently to avoid specifying what size each Digest's backing uses (addressing the whole concern about loss in usability).

I think that usability concern would also be solved with https://github.com/multiformats/rust-multihash/pull/156 wouldn't it? I'm in favor of that change, as I also think that people really shouldn't use the hashers directly.

mriise:
yep, that PR would do much of what I was thinking of doing. After playing around with const generic defaults a bit more I realized its not quite what I expected.

what i wanted to do was
```rust
struct Digest<const S: usize = DEFAULT_ALLOC_SIZE> ([u8; S]);

trait Hasher {
    type Digest: Digest;
}

impl<const S: usize> Hasher for Blake2<S> {
    type Digest = Blake2Digest<S>; // override Digest<DEFAULT_ALLOC> with Digest<S>
}
```
but instead `type Digest: Digest` seems to just turn into sugar for `type Digest: Digest<DEFAULT_ALLOC_SIZE>` meaning any impl generic over S will have conflicts, which makes the change essentially pointless.

As for resizing it can be done now with stable rust, but that is a minor change and isn't blocking this.

I would like to see #156 merged so we can reduce the amount of major releases :D

Stebalien (approved):
I have no problems merging this to master now, although I'd consider getting #156 and maybe #159 in first to batch up the rest of the breaking changes.

vmx:
@mriise @Stebalien Yes, the plan is to just merge this into master. I also want all other breaking changes that make sense in before doing a release.

@mriise #156 depends on this PR. So do I understand correctly that you're also OK merging this one?

mriise (approved):

PR #196, opened 2022-03-24
fix: remove the "identity" feature

### LINKED ISSUE #194: Remove `Code::Identity` support

Possible solutions:

Stebalien:
There are _always_ safer ways to compute an identity hash, and this feature can cause otherwise "safe" code to crash. With this feature enabled, `Code::try_from(codec)?.digest(input)` would crash if `codec` was 0 and `input` was more than 64 bytes.

Unfortunately, this was a cargo "feature" so a single dependency enabling it means it's enabled for all other crates in the build.

Users of this code migrate to `Multihash::wrap(0, digest)`, which returns an error if the digest is too large instead of panicing.

fixes #194

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,6 @@ serde-codec = ["serde", "serde-big-array"]
 
 blake2b = ["blake2b_simd"]
 blake2s = ["blake2s_simd"]
-identity = []
 sha1 = ["digest", "sha-1"]
 sha2 = ["digest", "sha-2"]
 sha3 = ["digest", "sha-3"]
--- a/src/multihash_impl.rs
+++ b/src/multihash_impl.rs
@@ -71,12 +71,6 @@ pub enum Code {
     #[cfg(feature = "blake3")]
     #[mh(code = 0x1e, hasher = crate::Blake3_256)]
     Blake3_256,
-
-    // The following hashes are not cryptographically secure hashes and are not enabled by default
-    /// Identity hash (max. 64 bytes)
-    #[cfg(feature = "identity")]
-    #[mh(code = 0x00, hasher = crate::IdentityHasher::<64>)]
-    Identity,
 }
 
 #[cfg(test)]


Stebalien:
Here's an example of a hack I've needed to prevent code that should "just work" from potentially crashing. I understand that downstream crates should generally use custom `Code` implementations, but that's not a reason to attach a foot-cannon to the default implementation:

https://github.com/filecoin-project/filecoin-ffi/blob/e81d302831f5dcd0c5b7b9415c353f8b43291b8c/rust/src/fvm/blockstore/fake.rs#L40-L43

vmx:
@mxinden one of the reasons this feature was originally introduces was rust-libp2p. Could you please check if it would be OK for the project if we would merge this change? I really want to prevent that projects stay on old versions/need to fork.

mxinden:
Thanks for the ping @vmx!

As far as I can tell our only usage of `Code::Identity` is in rust-libp2p's peer id handling:

https://github.com/libp2p/rust-libp2p/blob/6cc3b4ec52c922bfcf562a29b5805c3150e37c75/core/src/peer_id.rs#L55-L112

Do I understand correctly that we would either have to define our own `Code` enum, or use the magic constant (`0`) via `Multihash::wrap(0, digest)` and `multihash.code() == 0`?

Stebalien:
> Do I understand correctly that we would either have to define our own Code enum, or use the magic constant (0) via Multihash::wrap(0, digest) and multihash.code() == 0?

IMO, it's also reasonable to include some form of helper function _here_:

```rust
pub const IDENTITY_CODE: u64 = 0x0;
pub fn identity_hash<S>(data: &[0]) -> Multihash<S> {
    Multihash::wrap(IDENTITY_CODE, data)
}
```

vmx:
This feature got merged with https://github.com/multiformats/rust-multihash/pull/289, hence closing this one.

PR #198, opened 2022-03-30
Add Arbitrary implementation used for fuzzing

Kubuxu:
arbitrary::Arbitrary is a trait used for generation of arbitrary values
of structs used for fuzzing.

It differes from the quickcheck::Arbitrary by sourcing from a buffer
instead of Rng.

### DIFF
--- a/src/arb.rs
+++ b/src/arb.rs
@@ -1,13 +1,15 @@
-use quickcheck::{Arbitrary, Gen};
+use quickcheck::Gen;
 use rand::{
     distributions::{weighted::WeightedIndex, Distribution},
     Rng,
 };
 
+use arbitrary::{size_hint, Unstructured};
+
 use crate::MultihashGeneric;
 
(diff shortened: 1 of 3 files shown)


src/arb.rs
```diff
@@ -32,3 +34,47 @@ impl<const S: usize> Arbitrary for MultihashGeneric<S> {
         MultihashGeneric::wrap(code, &data[..size]).unwrap()
     }
 }
+
+impl<'a, const S: usize> AArbitrary<'a> for MultihashGeneric<S> {
```
Stebalien, line 38: maybe just scope this as `arbitrary::Arbitrary` and `quickcheck::Arbitrary`?

Stebalien (approved):
I'm can't speak to the distribution, but this should work.

Stebalien:
But you do need to run `cargo fmt --all`.

PR #199, opened 2022-03-30
[arb] make size_hint more precise

Kubuxu:
As we know that we won't need more more than S bytes to fill the buffer.


Kubuxu:
As it is an insignificant change. I'm merging it.

PR #201, opened 2022-04-01
Remove Identity hash from table, add helper function and examples of alternatives 

### LINKED ISSUE #194: Remove `Code::Identity` support

mriise:
related to #196 

closes #194

In order to allow for easier transition for downstream users, this adds a few ideas how to implement the identity hasher.

Undecided yet is about what to do with non-const sized identity hashers. Some users may be ok with allocating in order to have larger identity hashes. This is a fairly far edge-case which can be covered, but is debatable whether it should actually be actively supported by us.
(update): We won't be supporting alloc-dependent hashes in this crate for now unless significant desire exists to add them.

~~Another thing introduced is the `bytes` crate which can do fun zero-copy things. This part was mostly for me to become more familiar with the crate.~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,6 @@ quickcheck = "0.9.2"
 rand = "0.7.3"
 arbitrary = "1.1.0"
 multihash = { path = ".", features = ["sha1", "strobe"] }
-bytes = "1.1.0"
 
 [[bench]]
 name = "multihash"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,6 +62,7 @@ quickcheck = "0.9.2"
 rand = "0.7.3"
 arbitrary = "1.1.0"
 multihash = { path = ".", features = ["sha1", "strobe"] }
+bytes = "1.1.0"
 
 [[bench]]
 name = "multihash"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ serde-codec = ["serde", "serde-big-array"]
 
 blake2b = ["blake2b_simd"]
 blake2s = ["blake2s_simd"]
+# identity feature is depricated
 identity = []
 sha1 = ["digest", "sha-1"]
 sha2 = ["digest", "sha-2"]
--- a/examples/identity_hasher.rs
+++ b/examples/identity_hasher.rs
@@ -44,16 +44,19 @@ pub enum Code {
     Sha2_256,
 }
 
-#[allow(unused)]
 fn main() {
     // overwrite and trunk with code
 
     let src = b"hello world!";
     let ident_trunk = Code::IdentityTrunk.digest(src);
 
+    assert_eq!(ident_trunk.digest(), src);
+
     // input bigger than default table, but still const sized
     let big_src = b"Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus vehicula tempor magna quis egestas. Etiam quis rhoncus neque.";
-
+    
+    let truncated = Code::IdentityTrunk.digest(big_src);
+    assert_eq!(truncated.digest(), &big_src[..64]);
     //
     // Helper functions cannot be used with normal table if S != alloc_size, fancier const trait bounds in the future may allow for interop where S <= alloc_size
     //
@@ -73,5 +76,8 @@ fn main() {
 
     // makes use of the const sized input to infer the output size
     let big_arr_mh = identity_hash_arr(big_src);
+    assert_eq!(big_arr_mh.digest(), big_src);
+    // size must be specified
     let big_mh = identity_hash::<128>(big_src);
+    assert_eq!(big_mh.digest(), big_src);
 }
--- a/examples/identity_hasher.rs
+++ b/examples/identity_hasher.rs
@@ -1,5 +1,3 @@
-use bytes::{Buf, Bytes};
-
 use multihash::derive::Multihash;
 use multihash::{Error, Hasher, MultihashDigest, MultihashGeneric, Sha2_256};
 
@@ -37,49 +35,11 @@ impl<const S: usize> Hasher for IdentityTrunk<S> {
     }
 }
 
-/// Borrows and concats slices using Bytes crate
-#[derive(Debug, Default)]
-pub struct IdentityBorrow(Bytes);
-
-impl Hasher for IdentityBorrow {
-    fn update(&mut self, input: &[u8]) {
-        let new = (*self.0).chain(input);
-        self.0 = Bytes::copy_from_slice(new.chunk());
-    }
-    fn finalize(&mut self) -> &[u8] {
-        &self.0
-    }
-    fn reset(&mut self) {
-        *self = Self::default();
-    }
-}
-
-/// update allocates & appends, Alloc only
-#[derive(Default, Debug)]
-pub struct IdentityAlloc(Vec<u8>);
-
-impl Hasher for IdentityAlloc {
-    fn update(&mut self, input: &[u8]) {
-        self.0.extend(input);
-    }
-    fn finalize(&mut self) -> &[u8] {
-        &self.0
-    }
-    fn reset(&mut self) {
-        self.0.clear();
-    }
-}
-
 #[derive(Clone, Copy, Debug, Eq, Multihash, PartialEq)]
 #[mh(alloc_size = 64)]
 pub enum Code {
     #[mh(code = 0x00, hasher = IdentityTrunk::<64>)]
     IdentityTrunk,
-    // These will panic and if input > alloc_size, use alternate methods if you want to use bigger than alocated size
-    // #[mh(code = 0x02, hasher = IdentityBorrow)]
-    // IdentityBorrow,
-    // #[mh(code = 0x03, hasher = IdentityAlloc)]
-    // IdentityAlloc,
     #[mh(code = 0x12, hasher = Sha2_256)]
     Sha2_256,
 }
@@ -90,14 +50,15 @@ fn main() {
     let src = b"hello world!";
     let ident_trunk = Code::IdentityTrunk.digest(src);
 
-    // for bigger than default table, but still known smaller than a const
-
+    // input bigger than default table, but still const sized
     let big_src = b"Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus vehicula tempor magna quis egestas. Etiam quis rhoncus neque.";
 
+    //
     // Helper functions cannot be used with normal table if S != alloc_size, fancier const trait bounds in the future may allow for interop where S <= alloc_size
+    //
 
     pub const IDENTITY_CODE: u64 = 0x0;
-    // input must be const sized for this
+    // blind copy of the input array
     pub fn identity_hash_arr<const S: usize>(input: &[u8; S]) -> MultihashGeneric<S> {
         MultihashGeneric::wrap(IDENTITY_CODE, input).unwrap()
     }
@@ -109,6 +70,7 @@ fn main() {
         MultihashGeneric::wrap(IDENTITY_CODE, hasher.finalize()).unwrap()
     }
 
+    // makes use of the const sized input to infer the output size
     let big_arr_mh = identity_hash_arr(big_src);
     let big_mh = identity_hash::<128>(big_src);
 }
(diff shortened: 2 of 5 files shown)


examples/identity_hasher.rs
```diff
@@ -0,0 +1,114 @@
...
+#[derive(Clone, Copy, Debug, Eq, Multihash, PartialEq)]
+#[mh(alloc_size = 64)]
+pub enum Code {
+    #[mh(code = 0x00, hasher = IdentityTrunk::<64>)]
+    IdentityTrunk,
+    // These will panic and if input > alloc_size, use alternate methods if you want to use bigger than alocated size
+    // #[mh(code = 0x02, hasher = IdentityBorrow)]
+    // IdentityBorrow,
+    // #[mh(code = 0x03, hasher = IdentityAlloc)]
+    // IdentityAlloc,
```
vmx, line 82: In case you also want to add examples for those, you could use a code from the private area of the multicodec table (as you onviously can't use `0x00` for all of them: https://github.com/multiformats/multicodec/tree/f5dd49f35b26b447aa380e9a491f195fd49d912c#private-use-area

vmx (commented):
Thanks for this nice set of examples.

Would you then suggest that the upgrade path for users of the `identity` feature is, that they should use their own codec table?

mriise:
> Would you then suggest that the upgrade path for users of the `identity` feature is, that they should use their own codec table?

Yes, having a default can only cover one or maybe two use cases when it would be better for lib users to decide how they want to handle it.

examples/identity_hasher.rs
mriise: Borrow and Alloc examples were for me to play around with the concept of integrating. The alloc one specifically is funny since while it will hash the entirety of the input, it still gets shoved into the default sized array and crashes. this could be fixed by leaning into traits more and splitting `MultishashStack<S>` and `MultihashAlloc` into their own structs. However, good alloc support is not a good focus for this crate and would only apply to Identity hashing and maybe Blake3 with its extendable output.

I will the remove comments and references to them in the example, though if you want I can open a tracking issue about better alloc support to hopefully focus the attention of people that want it.

examples/identity_hasher.rs
```diff
@@ -0,0 +1,77 @@
...
+#[derive(Clone, Copy, Debug, Eq, Multihash, PartialEq)]
+#[mh(alloc_size = 64)]
+pub enum Code {
+    #[mh(code = 0x00, hasher = IdentityTrunk::<64>)]
+    IdentityTrunk,
+    #[mh(code = 0x12, hasher = Sha2_256)]
+    Sha2_256,
+}
+
+#[allow(unused)]
```
vmx, line 47: Instead of `#[allow(unused)]` you could add asserts. This way one would also get an idea what the output of those functions look like.

vmx (changes requested):
Please update the [modules top-level documentation](https://github.com/multiformats/rust-multihash/blob/master/src/lib.rs#L1-L50). It still mentions the identity hash. There should be some information about the identity hash on how to use it if you need it.

vmx:
@mxinden I'd like to get your approval before this one gets merged. As libp2p is one of the users of the identity hash.

mxinden:
Thanks for the ping @vmx. I am not quite sure I understand the consequences of this change for rust-libp2p.

Do I understand correctly that we would need to special-case the identity hash parsing from a `[u8]` here?

https://github.com/libp2p/rust-libp2p/blob/a4110a2b6939d5b460f11df6dd158308f517becf/core/src/peer_id.rs#L71-L75

vmx:
> Do I understand correctly that we would need to special-case the identity hash parsing from a `[u8]` here?

Yes that's correct. You'd also need to special case here: 
https://github.com/libp2p/rust-libp2p/blob/a4110a2b6939d5b460f11df6dd158308f517becf/core/src/peer_id.rs#L60-L66

Alternatively you could implement your own `Code` enum, which only includes SHA2-256 and the identity hash (which is also what the new example from this PR does).

mxinden:
That sounds reasonable to me. Thanks @vmx.

BigLep:
@mriise : are you able to incorporate comments here so this can be merged?

BigLep:
@mriise : are you able to incorporate comments here, or should we close?

src/lib.rs
```diff
@@ -92,3 +92,13 @@ pub use crate::hasher_impl::sha3::{Keccak224, Keccak256, Keccak384, Keccak512};
 pub use crate::hasher_impl::sha3::{Sha3_224, Sha3_256, Sha3_384, Sha3_512};
 #[cfg(feature = "strobe")]
 pub use crate::hasher_impl::strobe::{Strobe256, Strobe512, StrobeHasher};
+
+/// Code reserved for Identity "hash"
+pub const IDENTITY_CODE: u64 = 0x0;
```
thomaseizinger, line 97: This doesn't necessarily need to be public, can we make it crate-private?

src/multihash_impl.rs
```diff
@@ -83,12 +83,6 @@ pub enum Code {
     #[cfg(feature = "ripemd")]
     #[mh(code = 0x1055, hasher = crate::Ripemd320)]
     Ripemd320,
-
-    // The following hashes are not cryptographically secure hashes and are not enabled by default
-    /// Identity hash (max. 64 bytes)
-    #[cfg(feature = "identity")]
-    #[mh(code = 0x00, hasher = crate::IdentityHasher::<64>)]
-    Identity,
```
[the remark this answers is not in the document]
Stebalien: That's not possible, unfortunately. The footgun is that one might accidentally construct such a hasher, attempt to hash some data, then panic. This won't hit any "deprecated" warnings.

src/lib.rs
mriise: its still useful imo, e.g. `Multihash::wrap(multihash::IDENTITY_CODE, &bytes)`

examples/identity_hasher.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub fn identity_hash_arr<const S: usize>(input: &[u8; S]) -> MultihashGeneric<S> {
+        MultihashGeneric::wrap(IDENTITY_CODE, input).unwrap()
+    }
+
+    // input is truncated to S size
+    pub fn identity_hash<const S: usize>(input: &[u8]) -> MultihashGeneric<S> {
+        let mut hasher = IdentityTrunk::<S>::default();
+        hasher.update(input);
+        MultihashGeneric::wrap(IDENTITY_CODE, hasher.finalize()).unwrap()
+    }
```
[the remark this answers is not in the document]
mriise: The problem we have here is that these are both "valid" technically, the "what to do with too small of a buffer" is entirely based around the requirements of the system you are writing for. Most of the time I would go with identity_hash_arr since that forces users to fit data into a certain size before the library gets it.

These are examples anyway so just using the function defined in in lib.rs will be fine.

Cargo.toml
```diff
@@ -27,6 +27,7 @@ serde-codec = ["serde", "serde-big-array"]
 
 blake2b = ["blake2b_simd"]
 blake2s = ["blake2s_simd"]
+# identity feature is depricated
```
thomaseizinger, line 30:
```suggestion
# identity feature is deprecated
```

src/lib.rs
```diff
@@ -11,12 +11,11 @@
 //!
 //!  - `blake2b`: (default) Enable Blake2b hashers
 //!  - `blake2s`: (default) Enable Blake2s hashers
-//!  - `identity`: Enable the Identity hashers (using it is discouraged as it's not a hash function
-//!     in the sense that it produces a fixed sized output independent of the input size)
 //!  - `sha1`: Enable SHA-1 hasher
 //!  - `sha2`: (default) Enable SHA-2 hashers
 //!  - `sha3`: (default) Enable SHA-3 hashers
 //!  - `strobe`: Enable Strobe hashers
+//!  - `identity`: A depricated feature for identity hashes.
```
thomaseizinger, line 18:
```suggestion
//!  - `identity`: A deprecated feature for identity hashes.
```

src/multihash_impl.rs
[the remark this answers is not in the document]
thomaseizinger: I am not sure I understand. If the user uses this code, adding `#[deprecated]` to it will flag it right?

We can point the user to anything in the deprecation note, including explaining the footgun.

src/lib.rs
```diff
@@ -92,3 +91,13 @@ pub use crate::hasher_impl::sha3::{Keccak224, Keccak256, Keccak384, Keccak512};
...
+
+/// Code reserved for Identity "hash"
+pub const IDENTITY_CODE: u64 = 0x0;
+/// Helper for generating Identity hashes (context https://github.com/multiformats/rust-multihash/pull/196).
+/// Will error if `data.len() > S`
+///
+/// See examples for a few other approaches if this doesn't fit your application
+pub fn identity_hash<const S: usize>(data: impl AsRef<[u8]>) -> Result<MultihashGeneric<S>> {
+    MultihashGeneric::wrap(IDENTITY_CODE, data.as_ref())
+}
```
thomaseizinger, line 103: It seems inconsistent to me to provide only this here when we are showing two ways to do it in the example.

Why have this at all? Seems trivial to implement so I am not sure what we are gaining here. At the same time, we are adding an item to our public API which in the future might cause a breaking change.

src/multihash_impl.rs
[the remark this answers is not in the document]
thomaseizinger: To be clear, I am suggesting to add `#[deprecated]` to the enum variant.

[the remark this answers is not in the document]
Stebalien: The foot-gun is `Code::try_from(number).digest(...)`.

- If the identity feature gets enabled (e.g., by some other crate in the dependency tree).
- If `number` happens to be 0.

This will panic.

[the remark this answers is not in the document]
Stebalien: (assuming that digest is too long)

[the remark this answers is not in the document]
thomaseizinger: Right, now I get you. Thanks for explaining!

We can still add code that will automatically trigger a deprecation warning once the `identity` feature gets enabled. I did similar stuff in `rust-libp2p`.

Can push a PoC in a bit.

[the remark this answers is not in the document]
thomaseizinger: See https://github.com/multiformats/rust-multihash/pull/260.

src/lib.rs
thomaseizinger: I see that. I am just wary about extending the public API with new items that are not essential to be there. If we want to discourage the use of the identity hasher, it seems reasonable to have users declare that constant by themselves if they actually want to use it.

thomaseizinger (commented):
For the sake of making this a non-breaking change and working towards minimizing them in the future, can I ask that we:

- Not remove the `Identity` variant, but add `#[deprecated]` to it explicitly.
- Not add any new public API items. What we add here seems trivial to implement outside of the crate. It is my understanding that the `IdentityHasher` per se is not bad, users just need to be careful when using it. Having them write a bit more code seems like a code way of getting them to think what they are doing there.
- Apply https://github.com/multiformats/rust-multihash/pull/260 to warn users about the footgun in combination with the `identity` feature.

Thanks for considering :)

vmx:
> * Not add any new public API items.

I think that API comes from my request, when I said there needs to be a proper replacement for people that currently use it. I wanted a clean upgrade path. Though it can also be a well defined function in the (now existing ;) changelog, it doesn't have to be a public API (especially if we move the code table to a separate crate as it currently is discussed at https://github.com/multiformats/rust-multihash/issues/259).

thomaseizinger:
> > * Not add any new public API items.
> 
> I think that API comes from my request, when I said there needs to be a proper replacement for people that currently use it. I wanted a clean upgrade path. Though it can also be a well defined function in the (now existing ;) changelog, it doesn't have to be a public API (especially if we move the code table to a separate crate as it currently is discussed at https://github.com/multiformats/rust-multihash/issues/259).

I think a recommendation on how to deal with the identity hash is a good idea. I have no strong opinion whether it is an example, rustdoc comment, test,  changelog entry or something else :)

thomaseizinger:
There is https://github.com/multiformats/rust-multihash/pull/289 now which is up-to-date with latest master.

vmx:
As https://github.com/multiformats/rust-multihash/pull/289 was merged, I'll close this one.

PR #229, opened 2022-08-11
add `into_inner`

### LINKED ISSUE #220: Add getter for inner array
We have `fn digest(&self) -> &[u8` though it is helpful to also have `fn digest_arr(&self) -> [u8; S]` for `Multihash`

```rust 
fn digest_arr(&self) -> &[u8; S] {
    self.digest
}
```

mriise:
closes #220

### DIFF
--- a/src/multihash.rs
+++ b/src/multihash.rs
@@ -211,6 +211,26 @@ impl<const S: usize> Multihash<S> {
         mh.digest[..size].copy_from_slice(&self.digest[..size]);
         Ok(mh)
     }
+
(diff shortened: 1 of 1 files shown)


vmx (approved):
I'm still not a fan to make the API bigger, but there seems to be a use case for this.

BTW: I'm a huge fan of elaborate commit messages. Your comment in the source would make a great commit message next time.

vmx:
I've merged this one so that it's part of the next release.

PR #244, opened 2022-11-07
chore: update Cargo.lock

vmx:
`parity-scale-codec` and `cpufeatures` referred in the Cargo.lock file were pointing to yanked versions. Hence update.


Stebalien:
CI seems very sad, but I'm not sure why...

vmx:
CI is unhappy due to https://github.com/bkchr/proc-macro-crate/issues/26 :( I've restricted `proc_macro_crate` to the good version for now.

vmx:
The Nix build is failing, I think it's due to using an arbitrary nightly version. Do we care? (I know too little about Nix to fix it).

PR #254, opened 2022-12-05
fix: remove Nix support

### LINKED ISSUE #247: Nix build is currently broken
Currently the nix instructions are broken, likely due to using an arbitrary nightly version of Rust.

@Anderssorby @samuelburnham you were involved in setting it up. Could you please have a look?

vmx:
The Nix support isn't really maintained, hence remove it to reduce the maintenance burden.

The `Cargo.lock` file was only introduced for the Nix build, it's now removes as well as `multihash` is a library, which usually don't have lock files.

Closes #247.


samuelburnham:
Nix support is nice, but only if you have a dedicated maintainer who can keep track of the many different packages and overlays to integrate with Rust and Cargo. If not, Nix users can always write their own flake.nix or shell.nix.

Stebalien (approved):

PR #258, opened 2022-12-13
feat(multihash): update MSRV to 1.60.0 and use new syntax for cargo features

koushiro:
and some trivial things:

- use `sha1` instead of `sha-1` dependency
- update `strobe-rs` v0.7.0 ==> v0.8.1
- update `serde-big-array` v0.3.2 ==> v0.4.1
- update `multihash-derive` to edition 2021
- move the `resolver = "2"` to workspace

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,18 +1,19 @@
 [workspace]
 members = ["derive", "."]
+resolver = "2"
 
 [package]
 name = "multihash"
-description = "Implementation of the multihash format"
-repository = "https://github.com/multiformats/rust-multihash"
-keywords = ["multihash", "ipfs"]
 version = "0.18.0"
 authors = ["dignifiedquire <dignifiedquire@gmail.com>", "David Craven <david@craven.ch>", "Volker Mische <volker.mische@gmail.com>"]
+edition = "2021"
+rust-version = "1.60"
+description = "Implementation of the multihash format"
 license = "MIT"
 readme = "README.md"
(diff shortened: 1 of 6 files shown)


vmx (commented):
Is the `resolver = 2` needed in the workspace?

Moving to a new edition is usually a breaking change, hence it wasn't done for the derive crate yet, as I didn't want to do a breaking change release, just for upgrading the edition.

Also bumping the MSRV usually is a breaking change. I'd like to keep doing breaking releases to a minimum. Is there a way doing the crate updates without bumping the MSRV (or was bumping it the whole point?)?

koushiro:
> Is the `resolver = 2` needed in the workspace?

The resolver is a global setting for a [workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html), and the setting is ignored in dependencies. The setting is only honored for the top-level package of the workspace.

see https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html#details for details.

> Moving to a new edition is usually a breaking change, hence it wasn't done for the derive crate yet, as I didn't want to do a breaking change release, just for upgrading the edition.
> 
> Also bumping the MSRV usually is a breaking change. I'd like to keep doing breaking releases to a minimum. 

I know the upgrading the edition and MSRV are breaking changes, so you can merge this PR next time you need to upgrade the minor version.

thomaseizinger:
I think most of this happened in https://github.com/multiformats/rust-multihash/pull/272 now.

PR #263, opened 2022-12-29
Add cargo deny to CI

iamjpotts:
1. Locks the list of allowed licenses to the current set of dependency licenses
2. Surfaces two advisories as warnings (comments added to `deny.yml` for both)
3. Surfaces duplicate versions of the same dependency as warnings (comments added to `deny.yml` for both)

One of the advisories is for `atty` a dependency of `criterion` and tracked in an [issue in the criterion repo](https://github.com/bheisler/criterion.rs/pull/628).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,3 +96,28 @@ jobs:
 
       - name: Upload Code Coverage
         uses: codecov/codecov-action@v3
+
+  cargo-deny:
+    name: Cargo Deny
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Sources
+        uses: actions/checkout@v3
+
+      - name: Install Rust Toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+
+      - name: Install cargo-deny
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: cargo-deny
+
+      - name: Cargo Deny - Check
+        uses: actions-rs/cargo@v1
+        with:
+          command: deny
+          args: check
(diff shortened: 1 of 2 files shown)


iamjpotts:
> Thanks for your PR! I've just merged a PR that fixes the Clippy errors on CI. Could you please rebase your PR? I think it's a good idea and we should give it a try. If it turns out a too big burden, we can always remove it again. Also thanks for adding the exception for the current problems, so that we can start in a "green" state and work from there.

@vmx 

Rebased, and made a change to the workflow job to how it installs `cargo-deny`.

.github/workflows/build.yml
```diff
@@ -96,3 +96,28 @@ jobs:
 
       - name: Upload Code Coverage
         uses: codecov/codecov-action@v3
+
+  other:
+    name: Other
```
vmx, line 101: Could you please change it to something like `cargo-deny`, so that it's clearer on the CI UI, what is actually run.

iamjpotts: Done.

.github/workflows/build.yml
```diff
@@ -90,9 +90,34 @@ jobs:
           override: true
 
       - name: Generate Code Coverage
-        uses: actions-rs/tarpaulin@v0.1
+        uses: actions-rs/tarpaulin@v0.1.3
```
vmx, line 93: This change is not really needed. I thing `@0.1` should automatically pick up patch versions.

iamjpotts: That was to tickle the CI to see if a rerun would make the code coverage work.

vmx (approved):
Thanks for spending so much time on this!

PR #269, opened 2023-01-27
chore(deps): update proc-macro-crate to v1.3.0

koushiro:
close #268


koushiro:
waiting for #270 to be merged firstly

.github/workflows/build.yml
```diff
@@ -73,7 +73,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --no-default-features --workspace --target thumbv6m-none-eabi
+          args: --no-default-features --target thumbv6m-none-eabi
```
vmx, line 76: The whole workpace should work with no_std.

koushiro: Actually `proc-macro-crate` v1.3.0 introduces some new dependencies, and these deps do not support `no_std`.

```
➜  proc-macro-crate git:(master) cargo build --release --target thumbv6m-none-eabi
   Compiling autocfg v1.1.0
   Compiling memchr v2.5.0
   Compiling hashbrown v0.12.3
   Compiling toml_datetime v0.5.1
   Compiling once_cell v1.17.0
error[E0463]: can't find crate for `std`
  |
  = note: the `thumbv6m-none-eabi` target may not support the standard library
  = note: `std` is required by `once_cell` because it does not declare `#![no_std]`

error[E0463]: can't find crate for `std`
  |
  = note: the `thumbv6m-none-eabi` target may not support the standard library
  = note: `std` is required by `toml_datetime` because it does not declare `#![no_std]`

error[E0463]: can't find crate for `std`
 --> /Users/qinxuan/.cargo/registry/src/github.com-1ecc6299db9ec823/toml_datetime-0.5.1/src/datetime.rs:1:5
```

koushiro: see https://github.com/multiformats/rust-multihash/actions/runs/4025041425/jobs/6917756303 for details.

vmx: Then it needs to be fixed there or we can't upgrade.

koushiro: multihash-derive is a proc-macro library, we only care about whether the code it generates supports no_std.

vmx (approved):
Thanks!

PR #276, opened 2023-03-21
docs: tidy up changelog

thomaseizinger:


vmx:
> Fair enough. At least in HTML, multiple L1 headings is not common.

That's a good point. I checked, the `conventionalcommits` output is very similar to the current one, but with `h2` as headers. What about switching to that instead?

thomaseizinger:
> > Fair enough. At least in HTML, multiple L1 headings is not common.
> 
> That's a good point. I checked, the `conventionalcommits` output is very similar to the current one, but with `h2` as headers. What about switching to that instead?

Sounds good to me.

PR #277, opened 2023-03-21
fix: avoid possible panic in error handling code

thomaseizinger:
The `unsigned_varint::io::ReadError` is marked as `#[non_exhaustive]`. This means new variants could be added in patch versions. Panicking in the `_` branch is a big no-no, we need to properly handle new variants that could be added to this enum.


vmx:
No need to update the changelog, That's part of the release process.

thomaseizinger:
@vmx This is also ready for another review now.

vmx (approved):

Issue #282, opened 2023-03-27
Unsafe unwrap could cause an application to panic  


kevin-valerio:
https://github.com/multiformats/rust-multihash/blob/8e14b1919bb0f60daa63f0b070278a2963603790/src/multihash.rs#L368

The `unwrap()` here might cause some crash using the library if an error occurred. For chains using Substrate, this could cause critical severity issues if an attacker is able to craft a malicious payload in order to trigger the unwrap. 

You can find an image of a malicious payload that will cause a panic below.
![image](https://user-images.githubusercontent.com/24193167/227899155-bbe9ce57-1dc1-4f95-b8d5-a2e9cbeaee32.png)

Below is an example of a fix that could be used : 

```rust
   if decode::is_last(b[i]) {
      match decode::u64(&b[..=i]) {
        Ok(val) => return Ok(val.0),
        Err(_) => return Err(Error::VarIntDecodeError),
      }
    }
```

vmx:
Your example malicious payload has 11 elements, but the we decode only up to 10 elements, due to the buffer having a length of 10. Do you have an example that leads to a panic?

Though, even if a panic cannot be triggered, it might still make sense to change the code for clarity.

vmx:
I've tried:

```rust
    #[test]
    fn decode_overflow() {
        let data = [203, 155, 0, 0, 0, 5, 67];
        let result = read_u64(&data[..]);
        println!("result: {:?}", result);
    }
```

Just as your error message, it returns an error that it's not minimal, but it does not panic. What am I missing?

kevin-valerio:
> 

Put in your Cargo.toml

```
[dependencies]

multihash = { version = "0.18.0", git="https://github.com/multiformats/rust-multihash", default-features = false, features = [ "multihash-impl", "sha2"] }
```

Because this function will have a different behaviour if you have `std` in the features (because of `#[cfg(not(feature = "std"))]`), so ensure std is not on the features

mattheworris:
We have also run into panics on this line with a NotMinimal encoding error. We are also using a Substrate, no-std environment. 
Any chance this fix will get merged in the near future?

vmx:
> We have also run into panics on this line with a NotMinimal encoding error.

That is an error and not a panic, isn't it?

I cannot reproduce the panic. If you can reproduce the panic, please provide a unit test for it.

mattheworris:
The NotMinimal error triggers a panic when it hits unwrap().

[...]

PR #283, opened 2023-03-27
Refactoring unwrap and handling errors

kevin-valerio:
Please, see https://github.com/multiformats/rust-multihash/issues/282 for more informations

### DIFF
--- a/src/multihash.rs
+++ b/src/multihash.rs
@@ -357,20 +357,23 @@ pub(crate) use unsigned_varint::io::read_u64;
 /// Adapted from unsigned-varint's generated read_u64 function at
 /// https://github.com/paritytech/unsigned-varint/blob/master/src/io.rs
 #[cfg(not(feature = "std"))]
-pub(crate) fn read_u64<R: io::Read>(mut r: R) -> Result<u64, Error> {
-    use unsigned_varint::decode;
+pub fn read_u64(r: &mut ByteCursor) -> Result<u64, Error> {
     let mut b = varint_encode::u64_buffer();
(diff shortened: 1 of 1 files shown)


vmx (commented):
It would be great to have a test with the malicious payload, to make sure we won't have a regression in the future.

mattheworris:
> It would be great to have a test with the malicious payload, to make sure we won't have a regression in the future.

I attempted to create a test, but I could not get the test to work properly in a no-std environment.

kevin-valerio:
> > It would be great to have a test with the malicious payload, to make sure we won't have a regression in the future.
> 
> I attempted to create a test, but I could not get the test to work properly in a no-std environment.

See [that link](https://github.com/multiformats/rust-multihash/issues/282#issuecomment-1485130230), I wrote a test there with the good Cargo.toml config to reproduce the bug.

mattheworris:
> > > It would be great to have a test with the malicious payload, to make sure we won't have a regression in the future.
> > 
> > 
> > I attempted to create a test, but I could not get the test to work properly in a no-std environment.
> 
> See [that link](https://github.com/multiformats/rust-multihash/issues/282#issuecomment-1485130230), I wrote a test there with the good Cargo.toml config to reproduce the bug.

To clarify, I could not create a test that would work with `cargo test` in this repo. 
Regardless of what I tried (`--no-default-features` flags, etc.), `std` was always defined.

thomaseizinger:
Replaced by https://github.com/multiformats/rust-multihash/pull/291.

PR #289, opened 2023-04-11
feat(codetable): remove `identity` hasher

### LINKED ISSUE #194: Remove `Code::Identity` support

thomaseizinger:
This itself is not a breaking changes because `multihash-codetable` isn't released yet. We add an example to `multihash` on how to do identity hashing.

Resolves #194.
Replaces #196.
Replaces #201.
Replaces #260.

### DIFF
--- a/examples/identity.rs
+++ b/examples/identity.rs
@@ -1,0 +1,9 @@
+use multihash::Multihash;
+
+fn main() {
+    let valid_identity_hash = Multihash::<64>::wrap(0, b"foobar").unwrap();
+    let invalid_identity_hash = Multihash::<2>::wrap(0, b"foobar");
+
+    assert_eq!(valid_identity_hash.digest(), b"foobar");
+    assert_eq!(invalid_identity_hash.unwrap_err().to_string(), "Invalid multihash size 6.");
+}
--- a/examples/identity.rs
+++ b/examples/identity.rs
@@ -1,5 +1,3 @@
-use multihash::Multihash;
-
 //! An example for how to use the "identity" hash of [`Multihash`].
 //!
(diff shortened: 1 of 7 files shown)


examples/identity.rs
```diff
@@ -0,0 +1,22 @@
...
+//! [`Multihash::wrap`] returns an error in case the provided digest is too big for the available space.
+//! Make sure you construct a [`Multihash`] with a large enough buffer for your data.
+//!
+//! Typically, the way you want to use the "identity" hash is:
+//! 1. Check if your data is smaller than whatever buffer size you chose.
+//! 2. If yes, store the data inline.
+//! 3. If no, hash it make it fit into the provided buffer.
+
+fn main() {
+    let valid_identity_hash = Multihash::<64>::wrap(0, b"foobar").unwrap();
```
vmx, line 17: What about adding a constant like `const IDENTITY_CODE: u64 = 0`, to make it a bit clearer what the magic value `0` in the `wrap()` call means.

thomaseizinger: Sounds good. I'll link to the csv in the repository too.

mriise (approved):

thomaseizinger:
> @Stebalien as you requested the removal, would you mind to have a quick look?

It has been 6 weeks and we are currently drafting a new release of `rust-libp2p` in which we'd like to include the latest multihash changes. After all, I've been making them to harden the API and reduce the number of breaking changes in the ecosystem :)

Given the two above approvals, I am merging this. Also note that this concerns mostly the `multihash-codetable` crate and thus can be patched afterwards easily.

PR #291, opened 2023-04-12

### LINKED ISSUE #282: Unsafe unwrap could cause an application to panic

vmx:

Fixes #282.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -61,11 +61,11 @@ pub(crate) fn unsigned_variant_to_multihash_error(err: unsigned_varint::io::Read
 }
 
 #[cfg(not(feature = "std"))]
-impl From<unsigned_varint::decode::Error> for Error {
-    fn from(error: unsigned_varint::decode::Error) -> Self {
-        Error {
-            kind: Kind::Varint(error),
-        }
+pub(crate) fn unsigned_varint_decode_to_multihash_error(
+    err: unsigned_varint::decode::Error,
+) -> Error {
+    Error {
+        kind: Kind::Varint(err),
     }
 }
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -60,6 +60,15 @@ pub(crate) fn unsigned_variant_to_multihash_error(err: unsigned_varint::io::Read
     }
 }
 
+#[cfg(not(feature = "std"))]
+impl From<unsigned_varint::decode::Error> for Error {
+    fn from(error: unsigned_varint::decode::Error) -> Self {
+        Error {
+            kind: Kind::Varint(error),
+        }
+    }
+}
+
 pub(crate) fn io_to_multihash_error(err: io::Error) -> Error {
     Error {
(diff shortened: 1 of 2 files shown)


src/multihash.rs
```diff
@@ -307,7 +307,10 @@ pub(crate) fn read_u64<R: io::Read>(mut r: R) -> Result<u64, Error> {
         if n == 0 {
             return Err(Error::insufficient_varint_bytes());
         } else if decode::is_last(b[i]) {
-            return Ok(decode::u64(&b[..=i]).unwrap().0);
+            match decode::u64(&b[..=i]) {
+                Ok(decoded) => return Ok(decoded.0),
+                Err(_) => return Err(Error::varint_not_minimal()),
+            }
```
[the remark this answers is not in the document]
vmx: Thanks, that's even better!

src/error.rs
```diff
@@ -60,6 +60,15 @@ pub(crate) fn unsigned_variant_to_multihash_error(err: unsigned_varint::io::Read
     }
 }
 
+#[cfg(not(feature = "std"))]
+impl From<unsigned_varint::decode::Error> for Error {
```
vmx, line 64: @thomaseizinger I've implemented this with the `From` trait. Are there any reasons doing it with a function like the other conversions?

I'm even thinking about doing a PR switching the other ones to from traits as well, which seems more idiomatic to me.

thomaseizinger: The `From` trait implementation makes this part of the public API. Given that `unsigned_varint` is not yet `1.0`, I'd advise against it because it would mean that we have to also issue a new minor version if we update `unsigned_varint` to a new minor version.

thomaseizinger: It is more idiomatic but unfortunately a semver hazard because trait implementations don't stop at the crate boundary.

thomaseizinger: > Though I don't really like it.

I can relate. I also don't like the verbosity. Although I think the trade-off for being able update dependencies without cascading breaking changes up is definitely worth it.

src/multihash.rs
```diff
@@ -307,7 +307,9 @@ pub(crate) fn read_u64<R: io::Read>(mut r: R) -> Result<u64, Error> {
         if n == 0 {
             return Err(Error::insufficient_varint_bytes());
         } else if decode::is_last(b[i]) {
-            return Ok(decode::u64(&b[..=i]).unwrap().0);
+            return decode::u64(&b[..=i])
+                .map(|decoded| decoded.0)
+                .map_err(crate::error::unsigned_varint_decode_to_multihash_error);
```
thomaseizinger, line 312: A suggestion: https://github.com/multiformats/rust-multihash/pull/296

mxinden (approved):
Looks good to me. Thanks @vmx!

PR #292, opened 2023-04-12

### LINKED ISSUE #282: Unsafe unwrap could cause an application to panic

vmx:

Fixes #282.


vmx:
This PR is superseded by https://github.com/multiformats/rust-multihash/pull/293.

PR #293, opened 2023-04-13

### LINKED ISSUE #282: Unsafe unwrap could cause an application to panic

vmx:

Also some minor Clippy warnings were fixed.

Fixes #282.

### DIFF
--- a/src/multihash.rs
+++ b/src/multihash.rs
@@ -247,7 +247,8 @@ impl<const S: usize> Multihash<S> {
 }
 
 // Don't hash the whole allocated space, but just the actual digest
-#[allow(clippy::derive_hash_xor_eq)]
+#[allow(unknown_lints, renamed_and_removed_lints)]
+#[allow(clippy::derived_hash_with_manual_eq, clippy::derive_hash_xor_eq)]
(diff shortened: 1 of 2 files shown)


src/multihash.rs
```diff
@@ -362,7 +363,9 @@ pub(crate) fn read_u64<R: io::Read>(mut r: R) -> Result<u64, Error> {
         if n == 0 {
             return Err(Error::Varint(decode::Error::Insufficient));
         } else if decode::is_last(b[i]) {
-            return Ok(decode::u64(&b[..=i]).unwrap().0);
+            return decode::u64(&b[..=i])
+                .map(|decoded| decoded.0)
+                .map_err(Error::Varint);
```
mattheworris, line 368: I tested this change with our project and it fixes the panic, Thanks!

mattheworris (approved):
Thanks for the quick response.

mxinden (approved):
With my limited knowledge of `rust-multihash`, this looks good to me.

PR #295, opened 2023-04-14
fix(ci): switch to tarpaulin docker image

thomaseizinger:
This PR fixes the coverage job by switching to the tarpaulin docker container from the (unmaintained) Github action.


vmx (approved):
Thanks a lot! I was about to merge the PR removing the code coverage once the panic fix is merged. But this is even better!

PR #302, opened 2023-04-27
feat: Serde serialize Multihash in bytes representation

vmx:
Multihash has a defined bytes representation, serialize it as such one instead of deriving it from the internal representation. This way there is no leak of internal implementation details.

BREAKING CHANGE: The Serde serialization format changed

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,14 +22,13 @@ alloc = []
 arb = ["dep:quickcheck", "dep:rand", "dep:arbitrary"]
 scale-codec = ["dep:parity-scale-codec"]
 serde-codec = ["serde"] # Deprecated, don't use.
-serde = ["dep:serde", "dep:serde-big-array"]
+serde = ["dep:serde"]
 
 [dependencies]
 parity-scale-codec = { version = "3.0.0", default-features = false, features = ["derive"], optional = true }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -22,14 +22,13 @@ alloc = []
...
 scale-codec = ["dep:parity-scale-codec"]
 serde-codec = ["serde"] # Deprecated, don't use.
-serde = ["dep:serde", "dep:serde-big-array"]
+serde = ["dep:serde"]
 
 [dependencies]
 parity-scale-codec = { version = "3.0.0", default-features = false, features = ["derive"], optional = true }
 quickcheck = { version = "1.0.3", optional = true }
 rand = { version = "0.8.5", optional = true, features = ["small_rng"] }
 serde = { version = "1.0.116", optional = true, default-features = false, features = ["derive"] }
```
thomaseizinger, line 31:
```suggestion
serde = { version = "1.0.116", optional = true, default-features = false }
```

I think we can drop that feature now!

thomaseizinger (approved):
LGTM!

Cargo.toml
vmx: Good catch!

mxinden (approved):
:pray:

PR #303, opened 2023-04-27
Add job that verifies MSRV for `multihash`

thomaseizinger:


.github/workflows/build.yml
```diff
@@ -75,6 +75,26 @@ jobs:
...
+          MSRV=$(cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.name == "multihash") | .rust_version')
+          echo "MSRV=$MSRV" >> $GITHUB_ENV
+
+      - uses: dtolnay/rust-toolchain@v1
+        with:
+          toolchain: ${{ env.MSRV }}
+
+      - uses: Swatinem/rust-cache@v2
+
+      - run: cargo +"$MSRV" build --all-features --package multihash
```
vmx, line 96: Why no building the whole workspace? It should be good enough to just rely on the MSRV of the `multihash` crate.

vmx (commented):
That's so cool.

.github/workflows/build.yml
thomaseizinger: > Why no building the whole workspace? It should be good enough to just rely on the MSRV of the `multihash` crate.

We can set an MSRV for the whole workspace. I think it is good policy to consider breaking MSRV a breaking change.

The other crates have a lot more dependencies so I think being fine-grained here makes sense. I wouldn't want a breaking change in multihash just because some dependency in `multihash-codetable` decides on a new MSRV.

thomaseizinger: We can adjust the job to test the MSRV of each crate if you want.

Personally, I only care about `multihash` and none of the rest :)

vmx: > We can adjust the job to test the MSRV of each crate if you want.

Would be nice, but I'm also OK with keeping it to `multihash` only. It won't block merging it.

PR #305, opened 2023-04-29
Const generic default for `Multihash`

mriise:
Helps smooth over refactoring that expect `Multihash` to be a type alias for `Multihash<64>` already.

### DIFF
--- a/src/multihash.rs
+++ b/src/multihash.rs
@@ -37,7 +37,7 @@ use core2::io;
 #[cfg_attr(feature = "serde", derive(serde::Deserialize))]
 #[cfg_attr(feature = "serde", derive(serde::Serialize))]
 #[derive(Clone, Copy, Debug, Eq, Ord, PartialOrd)]
-pub struct Multihash<const S: usize> {
+pub struct Multihash<const S: usize = 64> {
     /// The code of the Multihash.
     code: u64,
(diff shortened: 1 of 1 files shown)


vmx (commented):
I don't have a strong opinion on that one. Hnece I'd like @thomaseizinger's input as he might prefer having this one break intentionally.

thomaseizinger:
Don't really mind either. I think it is more correct to not have a default because this data structure has no "knowledge" about how large the digests of hash functions are.

Also, we don't know what hash functions users want to use this with.

Defaults is what users will use. Popular hash functions today like sha256 only need 32bytes. We'd "waste" 32 bytes on the stack a lot of times then where users didn't read the docs super carefully to discover that there is a default here.

If you feel strongly, I am happy to include this but on the whole, I think it is better without.

vmx:
@mriise What do you think about if we do *not* add a default, but rather add something to the release upgrade documentation saying something like:

> The `Multihash` struct is now always generic over the size, if you want the behaviour from previous releases, you can define your own type such as `type Multihash = multihash::Multihash<64>`.

mriise:
I'm not terribly opinionated about it, although I will say most implementations i've seen have been 64 bytes just to be on the "safe" side of things rather than being 32 bytes. 

type aliasing works just about as well so we can go with that.

PR #315, opened 2023-05-31
chore: increase MSRV to 1.64

vmx:
`toml_datetime` v0.6.2 needs Rust 1.64 as minimum supported Rust version, hence also update to that one.

This is not considered a breaking change as we haven't specified a MSRV in previous releases.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ license = "MIT"
 readme = "README.md"
 documentation = "https://docs.rs/multihash/"
 edition = "2021"
-rust-version = "1.63"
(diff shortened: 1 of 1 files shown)


thomaseizinger:
> This is not considered a breaking change as we haven't specified a MSRV in previous releases.

Can we do so going forward? At least in its default state, i.e. without any features activated?

thomaseizinger (approved):

vmx:
> Can we do so going forward? At least in its default state, i.e. without any features activated?

Sure, that's the plan :)

PR #318, opened 2023-06-05
feat: don't activate features for MSRV check

### LINKED ISSUE #317: Only enforce MSRV if no features are activated
As discussed in https://github.com/multiformats/rust-multihash/issues/306, we should change our policy to only guarantee a MSRV when no features are activated.

This requires us to:

1. Update the documentation to reflect this policy
2. Change the CI job to not pass `--all-features`

Additionally, I think we should limit our MSRV guarantees to `multihash`. `multihash-codetable` and `multihash-derive` are unlikely to be deeply nested dependencies and thus MSRV isn't as critical.

thomaseizinger:
We only activate `std` by default and don't use any very recent std features. Other features will bring in dependencies where we don't control the MSRV policy. Thus, we don't guarantee a MSRV with features activated.

Resolves #317.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -93,7 +93,7 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
 
-      - run: cargo +"$MSRV" build --all-features --package multihash
+      - run: cargo +"$MSRV" build --package multihash
 
   coverage:
     name: Code Coverage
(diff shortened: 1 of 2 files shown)


mxinden (approved):
For what my review is worth, this sounds reasonable to me.

vmx (approved):

PR #319, opened 2023-06-05
docs: make README not terribly outdated

thomaseizinger:
Currently, the README is terribly outdated. It will likely get outdated again because the code samples in there are not type-checked.

This should do for now and hopefully unblock the `v0.19` release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,11 @@ You should set this to the maximum size of the digest you want to support.
 ```rust
 use multihash::Multihash;
 
+const SHA2_256: u64 = 0x12;
+
 fn main() {
-    let hash = Multihash::<64>::wrap(0x12, b"my digest");
-    println!("{:?}", hash);
+	let hash = Multihash::<64>::wrap(SHA2_256, b"my digest");
+	println!("{:?}", hash);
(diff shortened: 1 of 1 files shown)


mxinden (approved):
Again, with my limited knowledge.

README.md
```diff
@@ -37,29 +37,34 @@ MSRV 1.51.0 due to use of const generics
...
+`Multihash` uses const-generics to define the internal buffer size.
+You should set this to the maximum size of the digest you want to support.
+
 ```rust
-use multihash::{Code, MultihashDigest};
+use multihash::Multihash;
 
 fn main() {
-    let hash = Code::Sha2_256.digest(b"my hash");
+    let hash = Multihash::<64>::wrap(0x12, b"my digest");
```
vmx, line 49: Nitpick (won't prevent a merge). I prefer that we don't promote using magic numbers:

```rust
use multihash::Multihash;

const SHA2_256: u64 = 0x12;

fn main() {
    let hash = Multihash::<64>::wrap(SHA2_256, b"my digest");
    println!("{:?}", hash);
}
```

Issue #330, opened 2023-07-18
multihash const constructors unusable


aatifsyed:
[`Multihash`'s constructors are `const fn`](https://docs.rs/multihash/0.19.0/multihash/struct.Multihash.html#method.wrap), but they can't actually be used as such: 
```rust
use multihash::Multihash;

const _: Multihash<64> = match Multihash::wrap(
    0x16,
    &[
        0x16, 0x20, 0x64, 0x4b, 0xcc, 0x7e, 0x56, 0x43, 0x73, 0x04, 0x09, 0x99, 0xaa, 0xc8, 0x9e,
        0x76, 0x22, 0xf3, 0xca, 0x71, 0xfb, 0xa1, 0xd9, 0x72, 0xfd, 0x94, 0xa3, 0x1c, 0x3b, 0xfb,
        0xf2, 0x4e, 0x39, 0x38,
    ],
) {
    Ok(o) => o,
    Err(_) => panic!(),
};
```
Fails to compile:
```
error[E0493]: destructor of `Result<Multihash<64>, multihash::Error>` cannot be evaluated at compile-time
  --> src/lib.rs:3:32
   |
3  |   const _: Multihash<64> = match Multihash::wrap(
   |  ________________________________^
4  | |     0x16,
5  | |     &[
6  | |         0x16, 0x20, 0x64, 0x4b, 0xcc, 0x7e, 0x56, 0x43, 0x73, 0x04, 0x09, 0x99, 0xaa, 0xc8, 0x9e,
...  |
9  | |     ],
10 | | ) {
   | |_^ the destructor for this type cannot be evaluated in constants
...
13 |   };
   |   - value is dropped here

For more information about this error, try `rustc --explain E0493`.
```

This is presumably because the top level error type is non-copy/contains an [io::Error](https://doc.rust-lang.org/std/io/struct.Error.html) which cannot be `const`
https://github.com/multiformats/rust-multihash/blob/16e68d8203c787768688dab8aed4295a32f33bf0/src/error.rs#L42-L50

Here is the source for the constructor
https://github.com/multiformats/rust-multihash/blob/16e68d8203c787768688dab8aed4295a32f33bf0/src/multihash.rs#L59-L75

Fix is either:
- drop the illusion of `const` from the `wrap` constructor. This is a breaking change
- have `wrap` return a new, const-compatible `InvalidSize` error in its constructor. This is also a breaking change. This is my preference, and I'm happy to create a PR for such a change.

[...]

PR #331, opened 2023-07-19
feat: introduce `Multihash::new` which can be called from const contexts

aatifsyed:
Resolves https://github.com/multiformats/rust-multihash/issues/330.

### DIFF
--- a/src/multihash.rs
+++ b/src/multihash.rs
@@ -54,7 +54,41 @@ impl<const S: usize> Default for Multihash<S> {
     }
 }
 
+/// Returned from [`Multihash::new`]
+#[non_exhaustive]
+pub enum CreationError {
+    /// The digest exceeds the capacity of the multihash
+    DigestTooBig,
+}
+
 impl<const S: usize> Multihash<S> {
+    /// Wrap the digest in a multihash
+    /// ```rust
+    /// # use multihash::Multihash;
+    /// const MULTIHASH: Multihash<64> = match Multihash::new(0, &[]) {
+    ///     Ok(ok) => ok,
+    ///     Err(_) => panic!(),
+    /// };
+    /// ```
+    pub const fn new(code: u64, input_digest: &[u8]) -> Result<Self, CreationError> {
+        if input_digest.len() > S {
+            return Err(CreationError::DigestTooBig);
+        }
+
+        let size = input_digest.len();
+        let mut digest = [0; S];
+        let mut i = 0;
+        while i < size {
+            digest[i] = input_digest[i];
+            i += 1;
+        }
+        Ok(Self {
+            code,
+            size: size as u8,
+            digest,
+        })
+    }
+
     /// Wraps the digest in a multihash.
     pub const fn wrap(code: u64, input_digest: &[u8]) -> Result<Self, Error> {
         if input_digest.len() > S {
--- a/src/multihash.rs
+++ b/src/multihash.rs
@@ -104,6 +104,8 @@ impl<const S: usize> Multihash<S> {
     }
 
     /// Wraps the digest in a multihash.
+    #[deprecated]
+    #[doc(hidden)]
     pub const fn wrap(code: u64, input_digest: &[u8]) -> Result<Self, Error> {
         match Self::new(code, input_digest) {
             Ok(ok) => Ok(ok),
--- a/src/multihash.rs
+++ b/src/multihash.rs
@@ -54,29 +54,36 @@ impl<const S: usize> Default for Multihash<S> {
     }
 }
 
-/// Returned from [`Multihash::new`]
+/// Returned from [`Multihash::new`] if the digest cannot fit inside the multihash.
 #[derive(Debug)]
-#[non_exhaustive]
-pub enum CreationError {
-    /// The digest exceeds the capacity of the multihash
-    DigestTooBig,
-}
+pub struct DigestTooBig(usize);
 
-impl fmt::Display for CreationError {
+impl fmt::Display for DigestTooBig {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            CreationError::DigestTooBig => {
-                f.write_str("The digest exceeds the capacity of the multihash")
-            }
-        }
+        // Must match the error message for `crate::Error`
+        f.write_fmt(format_args!("Invalid multihash size {}.", self.0))
     }
 }
 
 #[cfg(feature = "std")]
-impl std::error::Error for CreationError {}
+impl std::error::Error for DigestTooBig {}
+
+// Allows implementations of `multihash_derive::MultihashDigest` to construct the
+// error required by that trait without using the deprecated `Multihash::wrap`
+//
+// (This includes the derive macro for that trait)
+impl From<DigestTooBig> for crate::Error {
+    fn from(value: DigestTooBig) -> Self {
+        Error::invalid_size(value.0 as _)
+    }
+}
 
 impl<const S: usize> Multihash<S> {
-    /// Wrap the digest in a multihash
+    /// Constructs a new [`Multihash`].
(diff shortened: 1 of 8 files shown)


src/multihash.rs
```diff
@@ -54,12 +54,41 @@ impl<const S: usize> Default for Multihash<S> {
     }
 }
 
+/// Returned from [`Multihash::new`]
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum CreationError {
+    /// The digest exceeds the capacity of the multihash
+    DigestTooBig,
+}
```
thomaseizinger, line 63: I think I'd prefer a struct instead of an enum with a single variant, plus we should print the same message as `Error::invalid_size`.

src/multihash.rs
```diff
@@ -54,12 +54,41 @@ impl<const S: usize> Default for Multihash<S> {
...
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for CreationError {}
+
 impl<const S: usize> Multihash<S> {
-    /// Wraps the digest in a multihash.
-    pub const fn wrap(code: u64, input_digest: &[u8]) -> Result<Self, Error> {
+    /// Wrap the digest in a multihash
```
thomaseizinger, line 79:
```suggestion
    /// Constructs a new [Multihash].
    ///
    /// The provided digest must be created from a hash-function with the registered code.
```

src/multihash.rs
```diff
@@ -54,12 +54,41 @@ impl<const S: usize> Default for Multihash<S> {
...
+    /// ```rust
+    /// # use multihash::Multihash;
+    /// const MULTIHASH: Multihash<64> = match Multihash::new(0, &[]) {
+    ///     Ok(ok) => ok,
+    ///     Err(_) => panic!(),
+    /// };
+    /// ```
```
thomaseizinger, line 86:
```suggestion
    /// # Example usage in `const`
    ///
    /// ```rust
    /// # use multihash::Multihash;
    /// const MULTIHASH: Multihash<64> = match Multihash::new(0, &[]) {
    ///     Ok(ok) => ok,
    ///     Err(_) => panic!(),
    /// };
    /// ```
```

thomaseizinger:
> Last use of `wrap` standing is

We should also change that to not generate code that is using deprecated functionality.

thomaseizinger (approved):
Thank you!

This is fine with me but I'd like more input from other maintainers. cc @vmx @mxinden

thomaseizinger:
Do we have other `const` functions that might be affected by the same problem? I wonder what a long-term solution could be here.

In the issue, you said that the problem is the `io::Error` in `Kind`? I tried previously to remove that one because it only appears in certain functions. Perhaps that is worth exploring because most functions don't actually return an `io::Error`.

aatifsyed:
> > Last use of `wrap` standing is
> 
> We should also change that to not generate code that is using deprecated functionality.

This is not possible without further changes to `Error`, which [has no public constructors](https://docs.rs/multihash/0.19.0/multihash/struct.Error.html). I think that's outside the scope of this PR. But, to outline the problem,
`Error` can only be created using methods like `wrap`.
It [must be created](https://docs.rs/multihash-derive/0.9.0/multihash_derive/trait.MultihashDigest.html#tymethod.wrap) to implement `MultihashDigest`.

Neither `rustc -Dwarnings` nor `clippy -Dclippy::pedantic` complain about the use of this method in derived code. I can add `#[automatically_derived]` or `#[allow(deprecated)]` if you would prefer.

As an aside, [the docs for `multihash-derive` are mangled](https://docs.rs/multihash-derive/0.9.0/multihash_derive/index.html)

aatifsyed:
> Do we have other `const` functions that might be affected by the same problem? I wonder what a long-term solution could be here.
> 
> In the issue, you said that the problem is the `io::Error` in `Kind`? I tried previously to remove that one because it only appears in certain functions. Perhaps that is worth exploring because most functions don't actually return an `io::Error`.

This would be a breaking change.

If you'd like my opinion, you should `impl From<Error> for std::io::Error`, and have io functions return and `io::Result`, which may be `InvalidData` with a `source = Some(YourErrorType)`

aatifsyed:
I've gone ahead and implemented the required changes:
- `DigestTooBig` now contains a usize so it can match the error message from `Error`. I think this is an odd choice - you're telling the caller the length of the slice, but they can already find that out because they retain slice ownership.
- Added a new constructor (`From::from`) for `Error` so that the derive macro doesn't use `wrap`

thomaseizinger:
> If you'd like my opinion, you should `impl From<YourErrorType> for std::io::Error`, and have io functions return and `io::Result`, which may be `InvalidData` with a `source = Some(YourErrorType)`, where `YourErrorType` is a slimmer `Error`

Good idea! Would you mind capturing this in an issue?

> ((I think that's fine, you're pre 1.0.0, and the API should be allowed to evolve past current limitations))

Yes-ish. `multihash` is a foundational crate of the libp2p ecosystem. It appears pretty much in all public APIs because `multiaddr` and `PeerId` depend on it. A breaking change is a breaking change _for every dependency_. We just released a breaking change a few weeks ago and I'd like to not go through that effort again for a couple of months unless really necessary.

mxinden (commented):
No objections from my end.

vmx:
> If you'd like my opinion, you should `impl From<YourErrorType> for std::io::Error`, and have io functions return and `io::Result`, which may be `InvalidData` with a `source = Some(YourErrorType)`, where `YourErrorType` is a slimmer `Error`

That sounds like a better solution. I would actually prefer that and keep the `wrap` name and just doing a breaking change. I agree that breaking changes are a hassle, but it shouldn't prevent us from going for a clean API. As this is very specific, also project (like libp2p) don't need to upgrade, they can just keep using the current version as long as they are happy with it.

I'd also be OK with just having the proper solution merged into master and just keep it there before doing a release, in case more breaking changes are uncovered.

thomaseizinger:
> > If you'd like my opinion, you should `impl From<YourErrorType> for std::io::Error`, and have io functions return and `io::Result`, which may be `InvalidData` with a `source = Some(YourErrorType)`, where `YourErrorType` is a slimmer `Error`
> 
> That sounds like a better solution. I would actually prefer that and keep the `wrap` name and just doing a breaking change. I agree that breaking changes are a hassle, but it shouldn't prevent us from going for a clean API. As this is very specific, also project (like libp2p) don't need to upgrade, they can just keep using the current version as long as they are happy with it.
> 
> I'd also be OK with just having the proper solution merged into master and just keep it there before doing a release, in case more breaking changes are uncovered.

The issue is that you will have to immediately make a release because I guess that @aatifsyed would want to use this right away.

Not upgrading to latest solution is at best a band-aid and means what we might have to do more work later by backporting PRs instead of just being able to merge them into master.

An alternative suggestion could be to temporarily introduce a `wrap_const` function and remove it again on the next breaking change.

aatifsyed:
Sorry for the week's silence, I was on vacation.

> because I guess that @aatifsyed would want to use this right away.

You're correct, _but_ as the rust-cid folks pointed-out[^1], there's a useable workaround for `Copy` types [^2]: 
```rust
const fn const_unwrap<T: Copy, E>(r: Result<T, E>) -> T {
    let v = match r {
        Ok(r) => r,
        Err(_) => panic!(),
    };
    mem::forget(r);
    v
}
```
[^1]: https://github.com/multiformats/rust-cid/issues/138#issuecomment-1646977310
[^2]: https://github.com/filecoin-project/builtin-actors/blob/f84baa92638f85845bebeed2399b83e9768a40bd/runtime/src/runtime/empty.rs#L8-L15

That means forest features are no longer blocked  - we can just use `const_unwrap()`.

I think a lot of the discussion above about error types/API design is valuable, and maybe could be addressed in #332, but this PR (or something like it) can wait until after that.

PR #337, opened 2023-08-25
fix: make Serde (de)serialization no_std compatible

### LINKED ISSUE #336: regression serde feature does not work with no-std
With the release of v0.19 we broke the ability to use Serde in no_std contexts. When you compile with:

```console
$ cargo build --no-default-features --features serde
   Compiling multihash v0.19.0 (/tmp/rust-multihash)
error[E0432]: unresolved import `std`
 --> src/serde.rs:3:5
  |
3 | use std::fmt;
  |     ^^^ use of undeclared crate or module `std`

error: cannot find macro `format` in this scope
  --> src/serde.rs:35:46
   |
35 |             .map_err(|err| de::Error::custom(format!("Failed to deserialize Multihash: {}", err)))
   |                                              ^^^^^^

error: cannot find macro `format` in this scope
  --> src/serde.rs:48:46
   |
48 |             .map_err(|err| de::Error::custom(format!("Failed to deserialize Multihash: {}", err)))
   |                                              ^^^^^^

error[E0599]: no method named `to_bytes` found for reference `&Multihash<SIZE>` in the current scope
  --> src/serde.rs:17:42
   |
17 |         serializer.serialize_bytes(&self.to_bytes())
   |                                          ^^^^^^^^ method not found in `&Multihash<SIZE>`

error[E0433]: failed to resolve: use of undeclared type `Vec`
  --> src/serde.rs:43:25
   |
43 |         let mut bytes = Vec::new();
   |                         ^^^ use of undeclared type `Vec`

Some errors have detailed explanations: E0432, E0433, E0599.
For more information about an error, try `rustc --explain E0432`.
error: could not compile `multihash` (lib) due to 5 previous errors
```

This is a regression as it works with v0.18.

vmx:
`multihash` v0.19 introduced a regression. It's no longer possible to compile with `no_std` and `serde` feature enabled. This commit fixes that regression.

Due to generic_const_exprs not being a thing yet, we need to use unsafe code to provide this functionality.

Fixes #336.

### DIFF
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -1,20 +1,76 @@
 //! Multihash Serde (de)serialization
 
-use std::fmt;
+use core::{fmt, ptr, slice};
 
 use serde::{
     de::{self, SeqAccess, Visitor},
-    Deserialize, Deserializer, Serialize, Serializer,
+    ser, Deserialize, Deserializer, Serialize, Serializer,
 };
 
 use crate::Multihash;
+
+/// The maximum serialization size of `code` is 9 bytes (a large varint encoded u64) and for `size`
+/// is 2 bytes  (a large varint encoded u8), this makes a total of 11 bytes.
+const MAXIMUM_PREFIX_SIZE: usize = 11;
+
+/// The is currently no way to allocate an array that is some constant size bigger then a given
+/// const generic. Once `generic_const_exprs` are a thing, this struct will no longer be needed.
+/// Until then we introduce a hack. We allocate a struct, which contains two independent arrays,
+/// which can be specified with const generics. We then treat the whole struct as a slice of
+/// continuous memory.
+#[repr(C, packed)]
+struct Buffer<const SIZE_FIRST: usize, const SIZE_SECOND: usize> {
+    first: [u8; SIZE_FIRST],
+    second: [u8; SIZE_SECOND],
+}
+
+#[allow(unsafe_code)]
+impl<const SIZE_FIRST: usize, const SIZE_SECOND: usize> Buffer<SIZE_FIRST, SIZE_SECOND> {
+    fn new() -> Self {
+        let buffer = Self {
+            first: [0; SIZE_FIRST],
+            second: [0; SIZE_SECOND],
+        };
+
+        // Make sure that the struct allocated continuous memory, as we exploit that fact with the
+        // `as_slice` and `as_mut_slice()` methods.
+        let start_first = ptr::addr_of!(buffer.first) as *const u8;
+        let start_second = ptr::addr_of!(buffer.second) as *const u8;
+        unsafe {
+            // This should never happen, hence it's OK to panic.
+            if start_second.offset_from(start_first) != SIZE_FIRST as isize {
+                panic!("alignment of the struct is wrong")
+            }
+        }
+
+        buffer
+    }
+
+    fn len(&self) -> usize {
+        SIZE_FIRST + SIZE_SECOND
+    }
+
+    fn as_slice(&self) -> &[u8] {
+        let start = ptr::addr_of!(self.first) as *const u8;
+        unsafe { slice::from_raw_parts(start, self.len()) }
+    }
+
+    fn as_mut_slice(&mut self) -> &mut [u8] {
+        let start = ptr::addr_of_mut!(self.first) as *mut u8;
+        unsafe { slice::from_raw_parts_mut(start, self.len()) }
+    }
+}
 
 impl<const SIZE: usize> Serialize for Multihash<SIZE> {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: Serializer,
     {
-        serializer.serialize_bytes(&self.to_bytes())
+        let mut buffer = Buffer::<MAXIMUM_PREFIX_SIZE, SIZE>::new();
+        let bytes_written = self
+            .write(buffer.as_mut_slice())
+            .map_err(|_| ser::Error::custom("Failed to serialize Multihash"))?;
+        serializer.serialize_bytes(&buffer.as_slice()[..bytes_written])
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/serde.rs
```diff
@@ -1,20 +1,76 @@
...
+        // Make sure that the struct allocated continuous memory, as we exploit that fact with the
+        // `as_slice` and `as_mut_slice()` methods.
+        let start_first = ptr::addr_of!(buffer.first) as *const u8;
+        let start_second = ptr::addr_of!(buffer.second) as *const u8;
+        unsafe {
+            // This should never happen, hence it's OK to panic.
+            if start_second.offset_from(start_first) != SIZE_FIRST as isize {
+                panic!("alignment of the struct is wrong")
+            }
+        }
```
Stebalien, line 44: This shouldn't be necessary as `packed` guarantees this.

src/serde.rs
```diff
@@ -1,20 +1,76 @@
...
+    }
+
+    fn as_slice(&self) -> &[u8] {
+        let start = ptr::addr_of!(self.first) as *const u8;
+        unsafe { slice::from_raw_parts(start, self.len()) }
+    }
+
+    fn as_mut_slice(&mut self) -> &mut [u8] {
+        let start = ptr::addr_of_mut!(self.first) as *mut u8;
+        unsafe { slice::from_raw_parts_mut(start, self.len()) }
```
Stebalien, line 60: You should be able to write this as:

```suggestion
        unsafe { slice::from_raw_parts_mut(self as *mut u8, self.len()) }
```

src/serde.rs
```diff
@@ -1,20 +1,76 @@
...
+            // This should never happen, hence it's OK to panic.
+            if start_second.offset_from(start_first) != SIZE_FIRST as isize {
+                panic!("alignment of the struct is wrong")
+            }
+        }
+
+        buffer
+    }
+
+    fn len(&self) -> usize {
```
Stebalien, line 49: I'd write this as `const`, just for perf reasons (shouldn't matter but...)

src/serde.rs
Stebalien: But you could consider writing a _test_ to assert this (but I'd avoid doing this at runtime).

src/serde.rs
Stebalien: Really, I'd just use `mem::size_of::<Self>()`.

src/serde.rs
Stebalien: Actually, you can [`static_assertions`](https://docs.rs/static_assertions/latest/static_assertions) to assert that `mem::size_of::<Self>() == SIZE_FIRST + SIZE_SECOND` at compile time.

Stebalien: You can also assert tat `Self` and `[u8]` have the same alignment.

Stebalien (approved):
This looks correct as-is, but see my comments.

src/serde.rs
vmx: I've tried

```rust
const_assert!(mem::size_of::<Self>() == SIZE_FIRST + SIZE_SECOND);
```

but sadly it cannot be compiled:

```console
> cargo build --no-default-features --features serde
error[E0401]: can't use generic parameters from outer function
  --> src/serde.rs:38:38
   |
29 | impl<const SIZE_FIRST: usize, const SIZE_SECOND: usize> Buffer<SIZE_FIRST, SIZE_SECOND> {
   | ---- `Self` type implicitly declared here, by this `impl`
...
38 |         const_assert!(mem::size_of::<Self>() == SIZE_FIRST + SIZE_SECOND);
   |                                      ^^^^
   |                                      |
   |                                      use of generic parameter from outer function
   |                                      use a type here instead

error[E0401]: can't use generic parameters from outer function
  --> src/serde.rs:38:49
   |
29 | impl<const SIZE_FIRST: usize, const SIZE_SECOND: usize> Buffer<SIZE_FIRST, SIZE_SECOND> {
   |            ---------- const parameter from outer function
...
38 |         const_assert!(mem::size_of::<Self>() == SIZE_FIRST + SIZE_SECOND);
   |                                                 ^^^^^^^^^^ use of generic parameter from outer function

error[E0401]: can't use generic parameters from outer function
  --> src/serde.rs:38:62
   |
29 | impl<const SIZE_FIRST: usize, const SIZE_SECOND: usize> Buffer<SIZE_FIRST, SIZE_SECOND> {
   |                                     ----------- const parameter from outer function
...
38 |         const_assert!(mem::size_of::<Self>() == SIZE_FIRST + SIZE_SECOND);
   |                                                              ^^^^^^^^^^^ use of generic parameter from outer function

For more information about this error, try `rustc --explain E0401`.
error: could not compile `multihash` (lib) due to 3 previous errors
```

So would you prefer if I remove that run-time check and write a test instead?

src/serde.rs
vmx: Nope:

```
error[E0606]: casting `&mut Buffer<SIZE_FIRST, SIZE_SECOND>` as `*mut u8` is invalid
  --> src/serde.rs:61:44
   |
61 |         unsafe { slice::from_raw_parts_mut(self as *mut u8, self.len()) }
   |                                            ^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0606`.
error: could not compile `multihash` (lib) due to previous error
```

src/serde.rs
vmx: I've moved it into a test. Though if anyone has an idea on how to assert it at compile-time, that would even be better.

Stebalien: Ah... yeah, generics. I... don't. But I think a test is probably enough.

src/serde.rs
Stebalien: Well, `self as *mut Buffer<_, _> as *mut u8`.

Stebalien (approved):

src/serde.rs
```diff
@@ -172,4 +220,20 @@ mod tests {
...
+        let buffer = Buffer::<SIZE_FIRST, SIZE_SECOND>::new();
+
+        // Make sure that the struct allocated continuous memory, as we exploit that fact with the
+        // `as_slice` and `as_mut_slice()` methods.
+        let start_first = ptr::addr_of!(buffer.first) as *const u8;
+        let start_second = ptr::addr_of!(buffer.second) as *const u8;
+        #[allow(unsafe_code)]
+        unsafe {
+            assert_eq!(start_second.offset_from(start_first), SIZE_FIRST as isize);
+        };
```
thomaseizinger, line 237: This is testing implementation details. Wouldn't it be better to just write to the slice via `as_mut_slice` and read from it again via `as_slice` and compare the results?

If that works without panicking, your unsafe code is correct.

vmx: Good point, I added another test.

src/serde.rs
vmx: I went for `self as *mut _ as _`.

thomaseizinger (approved):

PR #360, opened 2024-01-07
fix(codetable): automatically include digest dep when needed

bishopcheckmate:
Previously it was needed to manually enable the optional digest
dependency when using any of sha1,2,3, or ripemd features, and
it could only be found out via reading the sources or codetable
failed to build.

Also included some additional feature guards to clean up warnings
or errors that ocured when building without all features.


codetable/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "multihash-codetable"
 description = "Default multihash code-table with cryptographically secure hash implementations"
-version = "0.1.1"
+version = "0.1.2"
```
vmx, line 4: We do version bumps when releasing it. So please keep the version at 0.1.1.

vmx (commented):
Thanks a lot, that looks great! I've just left a minor comment about the version number.

PR #361, opened 2024-01-07
docs(codetable): expose all items with required features on docs.rs

bishopcheckmate:
This change affects only builds on docs.rs. Previously docs.rs built
multihash-codetable with only the default features enabled, thus all
the content of the crate could be inspected only when reading sources.
This change exposes all items from the crate on docs.rs and labels
them with what features they require.

### DIFF
--- a/codetable/Cargo.toml
+++ b/codetable/Cargo.toml
@@ -57,3 +57,7 @@ required-features = ["blake2b", "sha2"]
 name = "manual_mh"
 path = "examples/manual_mh.rs"
 required-features = ["sha2"]
+
+[package.metadata.docs.rs]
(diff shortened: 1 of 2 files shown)


bishopcheckmate:
the result will look like eg. in [tokio](https://docs.rs/tokio/latest/tokio/#modules)

bishopcheckmate:
I've bumped the version both here and in #360, however I think only one should come with version bump at the end. Here version bump is required just to rebuild the documentation

bishopcheckmate:
removed the version bump

vmx (approved):
Thanks for the improvements!

PR #366, opened 2024-06-26
Derive `Arbitrary` for `multihash_codetable::Code`

expede:
I found myself writing a newtype wrapper over `Code` to get an `impl proptest::Arbitrary`, and thought "hey, why don't I upstream this one-liner?"

I tried to keep to your libraries and conventions as much as possible (e.g. `arb`), but please let me know if I missed anything.

(Some of the automated checks are failing, but I'm 90% sure those failures "shouldn't" be related to this change. It's late here, so I'll double check tomorrow)

### DIFF
--- a/codetable/Cargo.toml
+++ b/codetable/Cargo.toml
@@ -11,6 +11,7 @@ edition = "2021"
 [features]
 default = ["std"]
 std = ["blake2b_simd?/std", "blake2s_simd?/std", "blake3?/std", "digest?/std", "sha1?/std", "sha2?/std", "sha3?/std", "strobe-rs?/std", "ripemd?/std", "multihash-derive/std", "core2/std"]
+arb = ["dep:arbitrary", "std"]
 sha1 = ["dep:digest", "dep:sha1"]
 sha2 = ["dep:digest", "dep:sha2"]
 sha3 = ["dep:digest", "dep:sha3"]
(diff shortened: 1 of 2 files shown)


vmx:
The CI failures are unrelated, i've opened https://github.com/multiformats/rust-multihash/pull/367 for a fix.

expede:
After #367, the tests got far enough to flag that there was a legitimate warning. However, I've now run into an edge case with the Cargo Hack action:

```
  50 | #[cfg_attr(feature = "arb", derive(arbitrary::Arbitrary))]
     |                                    ^^^^^^^^^^^^^^^^^^^^
     |                                    |
     |                                    unreachable call
     |                                    any code following this `match` expression is unreachable, as all arms diverge
```

I think this is a valid warning for the compiler to raise, _but it only happens if **none** of the hash algorithms are enabled_ (which should never happen in normal use). The most pragmatic approach here is to `allow` the unreachable code (since it's useless without any hash algos anyway), so I did that. Please let me know if you'd prefer a different strategy. I wish that I could just `allow` the one `derive`, but it seems to not pick that up unless it's on the entire module (perhaps because it pertains to a macro?)

vmx (approved):
Thanks for the PR and figuring all the details out. The workaround the warning is totally fine.

expede:
_Thank you!_

PR #371, opened 2024-09-30
Some maintenance 

vmx:
This is a PR that bundles some maintenance work. With all these changes, the dependencies are up to date and the CI is green. It doesn't contain any major changes.

This PR shouldn't be squashed when merged as it's really individual changes only bundles as they are so small.

### DIFF
--- a/derive-impl/Cargo.toml
+++ b/derive-impl/Cargo.toml
@@ -13,7 +13,7 @@ proc-macro = true
 [dependencies]
 proc-macro2 = { version = "1.0.24", features = ["span-locations"] }
 proc-macro-crate = "3.1.0"
-proc-macro-error = "1.0.4"
+proc-macro-error2 = "2.0.1"
 quote = "1.0.7"
 syn = "2.0.66"
 synstructure = "0.13.1"
--- a/codetable/Cargo.toml
+++ b/codetable/Cargo.toml
@@ -29,7 +29,7 @@ digest = { version = "0.10.1", default-features = false, optional = true }
 sha1 = { version = "0.10.5", default-features = false, optional = true }
 sha2 = { version = "0.10.0", default-features = false, optional = true }
 sha3 = { version = "0.10.0", default-features = false, optional = true }
-strobe-rs = { version = "0.8.1", default-features = false, optional = true }
+strobe-rs = { version = "0.10.0", default-features = false, optional = true }
 ripemd = { version = "0.1.1", default-features = false, optional = true }
 multihash-derive = { version = "0.9.0", path = "../derive", default-features = false }
 core2 = { version = "0.4.0", default-features = false }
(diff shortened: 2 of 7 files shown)


derive-impl/Cargo.toml
```diff
@@ -13,7 +13,7 @@ proc-macro = true
 [dependencies]
 proc-macro2 = { version = "1.0.24", features = ["span-locations"] }
 proc-macro-crate = "3.1.0"
-proc-macro-error = "1.0.4"
+proc-macro-error2 = "2.0.1"
```
Stebalien, line 16: This crate doesn't seem too popular. What's the issue with keeping the old one?

vmx: The reason is https://rustsec.org/advisories/RUSTSEC-2024-0370.html. And if you run something like cargo-deny (which we do on CI), it will complain about it. And I would expect that maybe someone will complain, hence it's easier to upgrade to that one. Another point is that crates are moving on to `syn` v2, so you would've less duplicated deps as `proc-macro-error` is still using v1.

Stebalien: I'd be more concerned about using a random crate than using a so-called "unmaintained" crate, from a supply-chain perspective at least.

vmx: The supply-chain perspective is a good point. I'll update that PR when I find time.

Stebalien: (I have to wonder if we could just vendor it)

nazar-pc: Looks like a legit developer to me :shrug: 

Advisory warnings are annoying for downstream users for sure, so it is worth fixing one way or another.

Stebalien: If you find it annoying, please complain to the rustsec project.

Stebalien: To expand on that, it's one thing if an unmaintained project has security issues piling up. But jumping to the latest hot fork is just as likely to cause problems as not. Declaring a project unmaintained without a clear replacement is literally just dangling a vulnerable part of the OSS supply chain in front of potential attackers and asking them to "jump in and take a bite".

Stebalien: > https://rustsec.org/advisories/RUSTSEC-2024-0370.html lists 3 replacements, including proc-macro-error2 used in this PR.

I'm talking about *clear* replacements. In this case, both the advisory and the suggestion to use `proc-macro-error2` were submitted by the author of `proc-macro-error2`.

The other alternatives were added later, but still aren't clear replacements.

> So overall I think it is fair to warn users that this crate is unmaintained and if they have any issues with it in the future, there will be no easy solution for them, which is even worse when dependency is indirect.

RUSTSEC is a security project and treats these advisories as security advisories. Sure, I get warning about unmaintained security critical bits _if_ there is a clear maintained replacement, but reports like this one just cause people to complain and are more likely to _cause_ security issues than they are to resolve them.

Stebalien: Ok, here's a version that simply drops the offending dependency: https://github.com/multiformats/rust-multihash/pull/372.

Stebalien:
Merged by rebase in #372.

PR #372, opened 2024-10-22
fix: remove the unmaintained proc-macro-error dependency

Stebalien:
We can already emit errors directly with `syn`, so we can just drop this dependency entirely.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,7 +100,7 @@ jobs:
     name: Code Coverage
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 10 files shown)


Stebalien:
Note: I've included the other commits from https://github.com/multiformats/rust-multihash/pull/371 to make CI happy.

vmx (approved):
Thanks @stebalien for looking into this more closely and just getting rid of `proc-macro-error`!

Stebalien:
Merged by rebase because there are multiple separate changes.

PR #387, opened 2025-09-02
chore: update dependencies

vmx:
Update all dependencies to their latest versions and adapt the code accordingly.


rvagg (approved):
maybe dependabot would be good in here

PR #407, opened 2026-04-14
Replace core2 with no_std_io2

cwlittle:
`core2` was yanked a couple of hours ago. This MR removes the dependency in favor of `no_std_io2`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,11 @@ description = "Implementation of the multihash format"
 repository = "https://github.com/multiformats/rust-multihash"
 keywords = ["multihash", "ipfs"]
 version = "0.19.3"
-authors = ["dignifiedquire <dignifiedquire@gmail.com>", "David Craven <david@craven.ch>", "Volker Mische <volker.mische@gmail.com>"]
+authors = [
+  "dignifiedquire <dignifiedquire@gmail.com>",
+  "David Craven <david@craven.ch>",
+  "Volker Mische <volker.mische@gmail.com>",
+]
 license = "MIT"
 readme = "README.md"
 documentation = "https://docs.rs/multihash/"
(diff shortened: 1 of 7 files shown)


LockedThread:
bro is doing god's work fixing all of the core2 usages

Cargo.toml
```diff
@@ -21,21 +25,25 @@ all-features = true
 [features]
 default = ["std"]
 std = ["unsigned-varint/std", "alloc"]
-alloc = ["core2/alloc"]
+alloc = ["no_std_io2/alloc"]
 arb = ["dep:quickcheck", "dep:rand", "dep:arbitrary"]
 scale-codec = ["dep:parity-scale-codec"]
-serde-codec = ["serde"] # Deprecated, don't use.
+serde-codec = ["serde"]                               # Deprecated, don't use.
```
athola: [nit] formatting seems a bit weird here, was this autoformatted this way?

athola (changes requested):
added minor nit comment, rest approved

vmx (approved):
Thanks everyone involved! This is such a clean change now. Also thanks for checking out an older version of `no_std_io2` to reduce the number of changes.

jxs (approved):

jxs:
@vmx can we release a patch version with this? So that we can unblock `rust-libp2p` builds: https://github.com/libp2p/rust-libp2p/issues/6383

derive/src/lib.rs
```diff
@@ -78,7 +78,7 @@ impl fmt::Display for UnsupportedCode {
     }
 }
 
-impl core2::error::Error for UnsupportedCode {}
+impl no_std_io2::error::Error for UnsupportedCode {}
```
elenaf9, line 81: Why not `core::error::Error`?

vmx: We have a very low MSRV and I didn't want to increase by that patch release.

PR #410, opened 2026-04-15
chore: bump MSRV and remove `no_std_io2` from `multihash-derive`

hanabi1224:
- Bump MSVR to `1.81` as `core::error::Error` was stablized in `1.81`: https://releases.rs/docs/1.81.0/#stabilized-apis
- Remove `no_std_io2` from `multihash-derive`
- (Upcoming: make `no_std_io2` optional in a subsequent PR)

### DIFF
--- a/codetable/Cargo.toml
+++ b/codetable/Cargo.toml
@@ -5,6 +5,7 @@ version = "0.2.1"
 repository = { workspace = true }
 license = { workspace = true }
 edition = { workspace = true }
+rust-version = "1.81"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
--- a/codetable/Cargo.toml
+++ b/codetable/Cargo.toml
@@ -5,7 +5,6 @@ version = "0.2.1"
 repository = { workspace = true }
 license = { workspace = true }
 edition = { workspace = true }
-rust-version = { workspace = true }
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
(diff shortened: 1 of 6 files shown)


codetable/Cargo.toml
```diff
@@ -2,9 +2,10 @@
 name = "multihash-codetable"
 description = "Default multihash code-table with cryptographically secure hash implementations"
 version = "0.2.1"
-repository = "https://github.com/multiformats/rust-multihash"
-license = "MIT"
-edition = "2021"
+repository = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+rust-version = { workspace = true }
```
vmx: Here and it's the same for the other traits. I'd like to keep the MSRV separate. The core `multihash` should be kept quite low as that's the most used one. `codetable` can be a bit more relaxed as people shouldn't use it anyway.

src/error.rs
```diff
@@ -1,7 +1,8 @@
+use core::error::Error as StdError;
```
vmx: I would replace the two occurrences of `StdError` with `core::error::Error`.

vmx (commented):
Those changes shouldn't break anything and given that 1.81 is already over 1.5 years old, I think we could release those even as a patch release. What do you think @hanabi1224?

src/error.rs
hanabi1224: Fixed.

codetable/Cargo.toml
hanabi1224: Fixed. (The MSRV CI job requires all crates to support 1.81 tho)

vmx: I should've been more specific. I would like them to be able to be independent also in the future. But for now all should be set to "1.81". But it could happen that e.g. some crate in `multihash-codetable` needs a more recent version, and the we can just upgrade that and not the core crate.

hanabi1224: Set `rust-version = "1.81"` for all creates

vmx (approved):

PR #411, opened 2026-04-16
Embed minimal required `no_std_io2` traits

hanabi1224:
This PR embedded minimal `no_std_io2` traits that are required by `multihash`. It should require a minor version bump

### DIFF
--- a/codetable/src/hasher_impl.rs
+++ b/codetable/src/hasher_impl.rs
@@ -6,15 +6,16 @@
 ))]
 macro_rules! derive_write {
     ($name:ident) => {
-        impl<const S: usize> no_std_io2::io::Write for $name<S> {
-            fn write(&mut self, buf: &[u8]) -> no_std_io2::io::Result<usize> {
+        #[cfg(feature = "std")]
+        impl<const S: usize> std::io::Write for $name<S> {
+            fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
                 use multihash_derive::Hasher as _;
 
                 self.update(buf);
                 Ok(buf.len())
             }
 
-            fn flush(&mut self) -> no_std_io2::io::Result<()> {
+            fn flush(&mut self) -> std::io::Result<()> {
                 Ok(())
             }
         }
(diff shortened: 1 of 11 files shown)


vmx:
As it's not as urgent as getting a quick fix out, other things came in between. I'll review it next week.

codetable/src/hasher_impl.rs
```diff
@@ -6,15 +6,16 @@
 ))]
 macro_rules! derive_write {
     ($name:ident) => {
-        impl<const S: usize> no_std_io2::io::Write for $name<S> {
-            fn write(&mut self, buf: &[u8]) -> no_std_io2::io::Result<usize> {
+        #[cfg(feature = "std")]
```
vmx, line 9: Wouldn't it make sense to just add something like this at the top:

```rust
#[cfg(feature = "std")]
use std::io;

#[cfg(not(feature = "std"))]
use crate::no_std_io as io;
```

vmx (commented):
I've pushed a new commit, as it was easier to do it that way than leaving a comment. A review from you on my changes would be great.

Though I've one comment left, where I don't understand why you made it `std` only.

codetable/src/hasher_impl.rs
hanabi1224: Fixed. 
(Did not implement the custom `Write` trait to not allow downstream crates to depend on it. But that might not be an issue.)
(Top-level use may not work when the macro is used in another file, using the full trait name instead)

vmx: > Did not implement the custom `Write` trait to not allow downstream crates to depend on it.

That's actually a good point. My thought was "just keep the current behaviour" without questioning it. I don't have much experience with no-std environments. I guess you would probably implement your own `Write`-like trait that fits your purpose. You wouldn't want to take a random, stripped down implementation from multihash.

@hanabi1224 it sounds like you've more experience with no-std. So sorry for the back and forth, but I think it makes sense to implement `Write` only for std. If if it turns out to be a problem for downstream users, they'll tell us.

hanabi1224: @vmx reverted.

vmx (approved):
Thanks for the PR and so quick responses!

PR #412, opened 2026-04-27
chore: prepare v0.19.5 release

vmx:
The plan would be to release a v0.19.5 version of the main `multihash` crate as it shouldn't really be a breaking change, it just changes internal implementations.

There is a large MSRV bump, but even Rust 1.81 is already over 1.5 years old. If someone really needs to support older versions, they can fix multihash to v0.19.4. If it turn out to be an issue for anyone, please open an issue.

@hanabi1224 As you can see my idea would be to release this as a patch release for `multihash`. Also patch releases for `multihash-derive` and `multihash-derive-impl` But release a minor for `multihash-codetable`, due to not implementing that custom `Write` for no-std. Does that sounds correct to you, or do you propose something else (you feedback would really be appreciated)?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## [0.19.5](https://github.com/multiformats/rust-multihash/compare/v0.19.4...v0.19.5) (2026-04-27)
+
+### Features
+
+* embed minimal required core2/no_std_io2 traits ([#411](https://github.com/multiformats/rust-multihash/issues/411)) ([90fab31](https://github.com/multiformats/rust-multihash/commit/90fab31ce86b5e87d281c6d718e053376852bf07))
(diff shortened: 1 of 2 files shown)


hanabi1224:
@vmx LGTM. Please note that I believe in theory, the change in `multihash` is breaking `cid` (only in `no_std` context). (`cid` requires the same `Write` trait from `multihash`. But we did patch release for `no_std_io2` transition that I believe was breaking `cid` as well, and we could follow the same process with a fast follow-up `cid` release).

hanabi1224 (approved):

vmx:
Very good points. Yes, let's do a `cid` patch release directly after and hope for then best :)

vmx:
I've changed my mind and also did a patch release for `multihash-codetable`, so that just everything get upgraded automatically to the embedded no-std implementation instead of relying on `no_std_io2` which can be seen as just a stop-gap.

PR #423, opened 2026-08-04
Improve CI

vmx:
Simplify things and make `cargo-deny` actually run on the whole workspace.


vmx:
> don't generated-pr and stale come from unified-ci / web3-bot?
I think so, let's see what happens (I don't really use those automated things for this repo).

